PR #7, opened 2015-03-28
Put module name in ini file instead of command line

takluyver:
@Carreau, what do you think of this? The module name is moved from a command line argument to inside the ini file. It assumes that the file is called `flit.ini`, though there's a command line option that can be used to specify a different path.

I need to fix docs and tests for this, and I'll probably change the code around a bit, but do you think this is the right way to go in terms of the interface?


Carreau:
Seam reasonable, and much simpler interface.

PR #9, opened 2015-03-30
[WIP] auto create flit.ini

Carreau:
```
~/d/flit (master) $ flit flit init
package name [flit] :
package license [BSD] :
home page [https://github.com/takluyver/flit] :
author name [Matthias Bussonnier] :
author email [bussonniermatthias@gmail.com] :
```

What do you think ? would this be usefull ?

### DIFF
--- a/flit/__init__.py
+++ b/flit/__init__.py
@@ -8,6 +8,7 @@ import pathlib
 import shutil
 import sys
 import zipfile
+import re
 
 from . import common
 from . import inifile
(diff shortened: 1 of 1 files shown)


takluyver:
Maybe, but I'd like to get the core stuff a bit more stable before going into that. I also wonder if this should be the job of something like [cookiecutter](https://github.com/audreyr/cookiecutter/).

PR #10, opened 2015-04-06
Add flit.init module

takluyver:
This can be run with: `python3 -m flit.init` for now.

@Carreau - this is similar to your PR #9. I'm still going back and forth on whether I want it or not, but I spent some time working on it on the plane.

The main difference from your design is that I didn't want to tie it to git; instead I use entries from last time you ran the tool as defaults. I've also designed it so it would be fairly easy to build a different interface for the same functionality - I'm looking at Urwid as one possibility.

### DIFF
--- a/flit/init.py
+++ b/flit/init.py
@@ -1,0 +1,178 @@
+import configparser
+import json
+import os
+from pathlib import Path
+import re
+import sys
+
+def get_data_dir():
+    """Get the directory path for flit user data files.
+    """
+    home = os.path.realpath(os.path.expanduser('~'))
+
+    if sys.platform == 'darwin':
+        d = Path(home, 'Library')
+    elif os.name == 'nt':
+        appdata = os.environ.get('APPDATA', None)
+        if appdata:
+            d = Path(appdata)
+        else:
+            d = Path(home, 'AppData', 'Roaming')
+    else:
+        # Linux, non-OS X Unix, AIX, etc.
+        xdg = os.environ.get("XDG_DATA_HOME", None)
+        d = Path(xdg) if xdg else Path(home, '.local/share')
+
+    return d / 'flit'
+
+def get_defaults():
+    try:
+        with (get_data_dir() / 'init_defaults.json').open() as f:
+            return json.load(f)
+    except FileNotFoundError:
+        return {}
+
+def store_defaults(d):
+    data_dir = get_data_dir()
+    try:
+        data_dir.mkdir(parents=True)
+    except FileExistsError:
+        pass
+    with (data_dir / 'init_defaults.json').open('w') as f:
+        json.dump(d, f, indent=2)
+
+license_names_to_classifiers = {
+    'BSD': 'License :: OSI Approved :: BSD License',
+    'GPL': 'License :: OSI Approved :: GPL License',
+    'Apache': 'License :: OSI Approved :: Apache License'
+}
+
+class IniterBase:
+    def __init__(self, directory='.'):
+        self.directory = Path(directory)
+        self.defaults = get_defaults()
+
+    def validate_email(self, s):
+        # Properly validating an email address is much more complex
+        return bool(re.match(r'.+@.+', s))
+
+    def guess_module_name(self):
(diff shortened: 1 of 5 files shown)


flit/init.py
```diff
@@ -0,0 +1,178 @@
...
+        return {}
+
+def store_defaults(d):
+    data_dir = get_data_dir()
+    try:
+        data_dir.mkdir(parents=True)
+    except FileExistsError:
+        pass
+    with (data_dir / 'init_defaults.json').open('w') as f:
+        json.dump(d, f, indent=2)
```
Carreau: mixing json conf file and ini file in same project seem not that consistent, but up to you.

flit/init.py
```diff
@@ -0,0 +1,178 @@
...
+
+license_names_to_classifiers = {
+    'BSD': 'License :: OSI Approved :: BSD License',
+    'GPL': 'License :: OSI Approved :: GPL License',
+    'Apache': 'License :: OSI Approved :: Apache License'
+}
```
Carreau: I can see MIT and public domain as well, but I guess it can be refined later.

flit/init.py
```diff
@@ -0,0 +1,178 @@
...
+class IniterBase:
+    def __init__(self, directory='.'):
+        self.directory = Path(directory)
+        self.defaults = get_defaults()
+
+    def validate_email(self, s):
+        # Properly validating an email address is much more complex
+        return bool(re.match(r'.+@.+', s))
+
+    def guess_module_name(self):
```
Carreau: Want to check here too that module_name is identifier ?

flit/init.py
takluyver: I never said I was consistent - or if I did, now I say I'm not ;-)

I prefer INI for files that humans read and write directly, JSON for when I don't expect the user to look at it normally.

flit/init.py
```diff
@@ -0,0 +1,178 @@
...
+        else:
+            return None
+    
+    def update_defaults(self, author, author_email, module, home_page, license):
+        new_defaults = {'author': author, 'author_email': author_email,
+                        'license': license}
+        name_chunk_pat = r'\b{}\b'.format(re.escape(module))
+        if re.search(name_chunk_pat, home_page):
+            new_defaults['home_page_template'] = \
+                re.sub(name_chunk_pat, '{modulename}', home_page, flags=re.I)
```
Carreau: I can see one problem here in some case, when project/org have same name like github.com/ipython/ipython, but I guess it's fine. We can refine with more guesser at another point.

flit/init.py
takluyver: Sure, this can be added to. I was thinking of following Github's [choosealicense.com](http://choosealicense.com/), which keeps it simple by only offering three options. I misremembered which three, though - they go for MIT instead of BSD.

flit/init.py
takluyver: Good point, I ought to do that.

flit/init.py
```diff
@@ -0,0 +1,178 @@
...
+        if any(new_defaults[k] != self.defaults.get(k) for k in new_defaults):
+            self.defaults.update(new_defaults)
+            store_defaults(self.defaults)
+
+
+class TerminalIniter(IniterBase):
+    def prompt_text(self, prompt, default, validator):
```
Carreau: This and next one could be class methods. Or function. Depends on wether you are considering reusing them.

flit/init.py
takluyver: Yeah, this definitely isn't going to be exactly right for all cases, but it will work if you're churning out projects at github.com/Carreau/(project)

Carreau:
How's jetlag ? 
Looks good otherwise.

flit/init.py
Carreau: Might have been BSD at some point, and they changed after [that](https://github.com/blog/1964-open-source-license-usage-on-github-com)

flit/init.py
Carreau: Yep I think that's enough for now :-)

PR #14, opened 2015-04-10
work on Register only flag.

Carreau:
Does not really belong in upload, or wheel, 

so might need to be splitted away. 

What do you think ?

### DIFF
--- a/flit/__init__.py
+++ b/flit/__init__.py
@@ -69,7 +69,7 @@ def main(argv=None):
             sys.exit(e.args[0])
     elif args.subcmd == 'register':
         from .upload import register
-        mod, meta = common.matadata_and_module_from_ini_file(args.ini_file)
+        mod, meta = common.metadata_and_module_from_ini_file(args.ini_file)
         register(meta)
     elif args.subcmd == 'init':
         from .init import TerminalIniter
--- a/flit/__init__.py
+++ b/flit/__init__.py
@@ -5,7 +5,6 @@ import pathlib
 import sys
 
 from . import common
-from . import inifile
 from .log import enable_colourful_output
 
 __version__ = '0.4'
@@ -47,6 +46,10 @@ def main(argv=None):
         help="Prepare flit.ini for a new package"
     )
 
+    subparsers.add_parser('register',
+            help="register a package on PyPI without uploading any files"
+    )
+
     args = ap.parse_args(argv)
 
     enable_colourful_output()
@@ -64,6 +67,10 @@ def main(argv=None):
             Installer(args.ini_file, user=args.user, symlink=args.symlink).install()
         except (common.NoDocstringError, common.NoVersionError) as e:
             sys.exit(e.args[0])
+    elif args.subcmd == 'register':
+        from .upload import register
+        mod, meta = common.matadata_and_module_from_ini_file(args.ini_file)
+        register(meta)
     elif args.subcmd == 'init':
         from .init import TerminalIniter
         TerminalIniter().initialise()
--- a/flit/common.py
+++ b/flit/common.py
@@ -158,3 +158,12 @@ def make_metadata(module, ini_info):
     md_dict.update(get_info_from_module(module))
     md_dict.update(ini_info['metadata'])
     return Metadata(md_dict)
+
+def metadata_and_module_from_ini_path(ini_path):
+    from . import inifile
+    ini_info = inifile.read_pkg_ini(ini_path)
+    module = Module(ini_info['module'],
+                                ini_path.parent)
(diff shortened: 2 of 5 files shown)


takluyver:
What's the need for it? I mostly see the separate register and upload steps as a PyPI wart, which I am deliberately hiding from the user by having upload try to register if necessary.

Carreau:
Grab the name of a package when you don't yet have code to publish (typically all the jupyter_ stuff). 

I can see one that wish to only have registerd package and no uploaded version yet, beyond master on github.

takluyver:
I guess that makes sense. So I'd probably do it as a separate command (`flit register`), and not document it, other than in the argparse help ('register to PyPI without uploading')

Carreau:
Updated.

flit/__init__.py
```diff
@@ -58,11 +62,14 @@ def main(argv=None):
     elif args.subcmd == 'install':
         from .install import Installer
         Installer(args.ini_file, user=args.user, symlink=args.symlink).install()
+    elif args.subcmd == 'register':
+        from .wheel import WheelBuilder
+        WheelBuilder(args.ini_file).register()
```
takluyver: This shouldn't be a WheelBuilder method, I don't think. It shouldn't be necessary to do any of the wheel stuff in order to register the package. I'd make a new method in upload that loads the ini and module info and then calls register(). I'm not quite sure what to call the new function, though. Maybe the new one should be 'register' and the existing one `_register_metadata()` or something.

flit/__init__.py
```diff
@@ -47,6 +47,10 @@ def main(argv=None):
         help="Prepare flit.ini for a new package"
     )
 
+    parser_register = subparsers.add_parser('register',
+            help="register a package on PyPI"
```
takluyver: I'd like this to specifically say "...without uploading any files", to hint that that's the only case where you need it.

flit/__init__.py
```diff
@@ -58,11 +62,14 @@ def main(argv=None):
...
         Installer(args.ini_file, user=args.user, symlink=args.symlink).install()
+    elif args.subcmd == 'register':
+        from .wheel import WheelBuilder
+        WheelBuilder(args.ini_file).register()
     elif args.subcmd == 'init':
         from .init import TerminalIniter
         TerminalIniter().initialise()
     else:
-        sys.exit('No command specified')
+        sys.exit('No command specified, use --help for more informations')
```
takluyver: I wonder if argparse has a method to print the help info - rather than telling people to do `--help`, we could just act as if they had. No need for that to be in this PR, just a thought.

Carreau:
> This shouldn't be a WheelBuilder method, I don't think. It shouldn't be necessary to do any of the wheel stuff in order to register the package. I'd make a new method in upload that loads the ini and module info and then calls register(). I'm not quite sure what to call the new function, though. Maybe the new one should be 'register' and the existing one _register_metadata() or something.


Made a base class that load ini and set `self.metadata` plus one intermediate `Registerer` that can register, from which WheelBuilder  inherit.

thought ?

takluyver:
I'm inclined to do it with plain functions for now - I often find it easier to build APIs as functions first and then only refactor into classes once my ideas crystallise a bit. The hierarchy of Base -> Registerer -> WheelBuilder doesn't feel quite right to me.

takluyver:
Sorry if it feels like I'm being quite picky about code style - this is still my baby at the moment.

Carreau:
One more try.

flit/__init__.py
```diff
@@ -64,6 +67,10 @@ def main(argv=None):
             Installer(args.ini_file, user=args.user, symlink=args.symlink).install()
         except (common.NoDocstringError, common.NoVersionError) as e:
             sys.exit(e.args[0])
+    elif args.subcmd == 'register':
+        from .upload import register
+        mod, meta = common.matadata_and_module_from_ini_file(args.ini_file)
```
takluyver: s/matadata/metadata/

takluyver:
Other than that, I think this looks good.

Carreau:
Nice catch, fixed.

Should find a way top test agains testpypi.

takluyver:
There are at least a few libraries out there to mock requests so it doesn't actually make http requests - I suspect that's the way to do it. I don't know which are best, though.

Carreau:
I'm looking at how pip is doing it. but apparently it does not. It seem like it **should** be easy to pip-intall test-pypi to have a local fake server. I guess not many packages do that.

takluyver:
I think the pypi code itself requires a Postgres database running, so it's probably not pip installable. Hopefully when warehouse is the official thing, it will have a simpler API that we can easily test against.

PR #16, opened 2015-04-10
Better error message for missing docstring and version.

Carreau:
Did that roughly in the bart. 

Was not remembering how to print Error message only cleanly.


takluyver:
Thanks, that looks reasonable. Can you add the same try/except for install as well, since that also gets the version number?

Also, btw, you can `sys.exit("message")` to print a message and exit with code 1.

Carreau:
done.

takluyver:
Thanks!

takluyver:
Congratulations on getting your own [contributor graph](https://github.com/takluyver/flit/graphs/contributors)

PR #19, opened 2015-04-14
Update docs on flit init

Carreau:
Nothing fancy.


takluyver:
Thanks, but I'd prefer to do this slightly differently. I want the README to still be clear about what is actually used, and just mention `flit init` as a shortcut - I don't want to hide the ini file.

PR #22, opened 2015-04-14
Print config dir, version, and exit successfully.

Carreau:
Strait forward, but maybe you don't want it.


takluyver:
--version makes sense, but I think argparse has an easier way to do that specifically.

I'm less keen on --data-dir, because the user shouldn't need to know where that is. We can add a `log.debug()` that prints the location when it's used, perhaps. For that matter, a `--debug` flag to set the logging level to DEBUG would be useful.

takluyver:
Added --version in 45019c2e8e57a1e09e81f2be9f3fd4d97dc19993

PR #27, opened 2015-04-24
fix debug format

Carreau:
This still complain for me, it seem to take a format string, like info, warning, message...etc.


takluyver:
Oh, yes, debug is just another level. I thought we fixed that before. Strange.

Carreau:
no we just switched from log to debug befor.

PR #32, opened 2015-05-17
Start working on validating rst

Carreau:


flit.ini
```diff
@@ -3,7 +3,7 @@ module=flit
 author=Thomas Kluyver
 author-email=thomas@kluyver.me.uk
 home-page=https://github.com/takluyver/flit
-requires=requests
+requires=requests,docutils
```
takluyver: These should go one per line, like classifiers, I believe.

Carreau: Done.

Carreau:
```
$ python3 bootstrap_dev.py
  File "bootstrap_dev.py", line 14, in <module>
    Installer(p, symlink=True).install()
  File "/Users/bussonniermatthias/dev/flit/flit/install.py", line 55, in __init__
    self.ini_info = inifile.read_pkg_ini(ini_path)
  File "/Users/bussonniermatthias/dev/flit/flit/inifile.py", line 124, in read_pkg_ini
    raise ValueError("The file description seem to not be valid rst for PyPI")
ValueError: The file description seem to not be valid rst for PyPI


~/d/flit (validate-rst) $ python3 bootstrap_dev.py
Symlinking flit -> /Users/bussonniermatthias/Library/Python/3.4/lib/python/site-packages/flit
```

Carreau:
of interest : https://pypi.python.org/pypi/restview and https://github.com/msabramo/setuptools-markdown

flit/inifile.py
```diff
@@ -117,7 +117,12 @@ def read_pkg_ini(path):
     if 'description-file' in md_sect:
         description_file = path.parent / md_sect.pop('description-file')
         with description_file.open() as f:
-            md_dict['description'] = f.read()
+            raw_desc =  f.read()
+        from readme.rst import render
+        _, ok = render(raw_desc)
+        if not ok:
+            raise ValueError("The file description seem to not be valid rst for PyPI")
```
takluyver: Let's make this a `log.warn()`, not an exception:

1. Descriptions don't *have* to be rst - plain text is valid, that's why it silently falls back to unrendered text if there's an error producing rst.
2. This shouldn't prevent it from installing, where the description doesn't matter.

takluyver: Also, it would be good to indicate *why* it's not valid rst. I think the code from distutils gave a list of error messages. Actually, I think I'd rather copy the distutils code than create unnecessary dependencies on bleach and pygments (though they're a fitting pairing...) through readme.

Carreau: > Descriptions don't have to be rst - plain text is valid, that's why it silently falls back to unrendered text if there's an error producing rst.

Plain text would mostly not raise errors, but ok to log.warn.



> Also, it would be good to indicate why it's not valid rst. I think the code from distutils gave a list of error messages. Actually, I think I'd rather copy the distutils code than create unnecessary dependencies on bleach and pygments (though they're a fitting pairing...) through readme.

I understand, though I don't want to get into the buisness of replicating pipy logic in here. I can vendorize [readme/rst.py](https://github.com/pypa/readme/blob/master/readme/rst.py) that does not seem to require any of bleach or pygments.

flit/inifile.py
```diff
@@ -117,7 +120,14 @@ def read_pkg_ini(path):
...
         description_file = path.parent / md_sect.pop('description-file')
         with description_file.open() as f:
-            md_dict['description'] = f.read()
+            raw_desc =  f.read()
+        stream = io.StringIO()
+        _, ok = render(raw_desc, stream)
+        stream.seek(0)
+        if not ok:
+            log.warn("The file description seem to not be valid rst for PyPI it will be interpreted as plain text")
+            log.warn(stream.read())
```
takluyver: If you use `stream.getvalue()`, there's no need for the seek above.

flit/inifile.py
```diff
@@ -117,7 +120,14 @@ def read_pkg_ini(path):
     if 'description-file' in md_sect:
         description_file = path.parent / md_sect.pop('description-file')
         with description_file.open() as f:
-            md_dict['description'] = f.read()
+            raw_desc =  f.read()
+        stream = io.StringIO()
+        _, ok = render(raw_desc, stream)
+        stream.seek(0)
+        if not ok:
+            log.warn("The file description seem to not be valid rst for PyPI it will be interpreted as plain text")
```
takluyver: * seem -> seems
* to not be -> not to be
* some punctuation after PyPI - I'd use a semicolon

flit/inifile.py
Carreau: Ha, I was looking for it. Thanks.

Carreau:
`Readme/rst.py` "vendorized", requirement on docutils, + use `log.warn`.

takluyver:
Thanks :-)

PR #35, opened 2015-06-01
Allow markdown description !

Carreau:
OK, I know you don't want that, just that it works :-) 

Use pandoc to convert `Readme.md` to a rst description. [it works](https://pypi.python.org/pypi/SPQR) .

Feel free to close :-)


takluyver:
Hidden bonus feature!

PR #37, opened 2015-06-14
Suggest closest matching metadata key

takluyver:
This gives an error message like `flit.inifile.ConfigError: Unrecognised metadata key: require (did you mean 'requires'?)`

Ping @Carreau


Carreau:
+1.

PR #42, opened 2015-07-07
`flit install` installs requirements

### LINKED ISSUE #11: flit dont' like when modules have no docs.
```
$flit wheel
Fetching list of valid trove classifiers                          I-flit.inifile
Starting new HTTPS connection (1): pypi.python.org  I-requests.packages.urllib3.connectionpool
  File "/usr/local/bin/flit", line 11, in <module>
    sys.exit(main())
  File "/Users/bussonniermatthias/Library/Python/3.5/lib/python/site-packages/flit/__init__.py", line 57, in main
    verify_metadata=args.verify_metadata).build()
  File "/Users/bussonniermatthias/Library/Python/3.5/lib/python/site-packages/flit/wheel.py", line 31, in __init__
    self.metadata = common.make_metadata(self.module, self.ini_info)
  File "/Users/bussonniermatthias/Library/Python/3.5/lib/python/site-packages/flit/common.py", line 148, in make_metadata
    md_dict.update(get_info_from_module(module))
  File "/Users/bussonniermatthias/Library/Python/3.5/lib/python/site-packages/flit/common.py", line 38, in get_info_from_module
    docstring_lines = m.__doc__.splitlines()
AttributeError: 'NoneType' object has no attribute 'splitlines'
```

### LINKED ISSUE #41: flit install should install dependencies
The requires field should be a valid [requirements.txt](https://pip.pypa.io/en/latest/reference/pip_install.html#requirements-file-format) file, so we can shell out to pip to install these.

Suggestion from @minrk

minrk:
with a temporary requirements.txt build from metadata.requires_dist

closes #41

### DIFF
--- a/flit/install.py
+++ b/flit/install.py
@@ -7,6 +7,8 @@ import pathlib
 import shutil
 import site
 import sys
+import tempfile
+from subprocess import check_call
 
 from . import common
 from . import inifile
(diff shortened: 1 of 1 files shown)


minrk:
This should handle proper Requires-Dist format now.

flit/install.py
```diff
@@ -22,6 +25,32 @@
...
+    """
+    parts = requires_dist.split(';', 1) + ['']
+    distvers = parts[0].strip()
+    mark = parts[1].strip()
+    distvers = re.sub(_EQEQ, r"\1==\2\3", distvers)
+    distvers = distvers.replace('(', '').replace(')', '')
+    if not mark:
+        return distvers
+    # if an environment marker is present, check it:
+    from _markerlib import compile as compile_marker
```
takluyver: This is part of setuptools, so I don't want to import it. It should be OK to leave environment markers alone for requirements.txt and let pip handle them.

I think I'm going to write my own version of this function. It's taking me far more effort than it should to verify that it correctly does a relatively simple thing.

minrk: Rewritten a little to be simpler and avoid importing markerlib

takluyver:
Thanks, I think this looks good now.

PR #44, opened 2015-07-11
Crazy idea: specify flit configuration without `.ini` files

cjrh:
Hi @takluyver  

As discussed on twitter, here is a quick stab at what it might look like to have the metadata inside the code, rather than additional ini files. (Just for discussion obviously; not expecting you to merge this!)

I used the tests to make and test the ideas.  In this fork, there are no `.ini` files in the tests folder and **all tests pass**.  Thank you for taking the time to make good tests!

In short, the new config looks like this (e.g., `module1.py`):

```python
"""Example module"""

from flit.inifile import flit_config

__version__ = '0.1'
__FLIT__ = flit_config(
        version=__version__,
        module='module1',
        author='Sir Robin',
        author_email='robin@camelot.uk',
        home_page='http://github.com/sirrobin/module1'
        )
```

For packages, as you will see, this stuff goes into the `__init__.py` file.  I first played around with using all dunder vars, like `__AUTHOR_EMAIL__` and so on, but I quickly began to dislike it.

What's nice about doing it like *this* is that if your editor/ide has completion functionality, you get a lovely signature for the `flit_config()` function that shows you which options are mandatory including, potentially, a very useful docstring with additional info.  Here I made the non-required args all `**kwargs`, but in principle you could put more of them in the function signature for `flit_config()`, to help people know what to put in.  With setuptools you need to read a ton of stuff to know how to make a package; so much better to let your editor give it to you.

Anyway, hope you find it interesting.  It was a lot of fun.


takluyver:
Thanks for investigating this - it's an impressive amount of changes in a very short space of time.

A few downsides have occurred to me since we discussed this on twitter, though I'm still weighing up the benefits as well.

- The module name would have to go into the command line interface again, to tell it what to work with. This was how I did it very early in flit development (I started with .ini files inside the packages themselves), and when I switched to the current scheme of `flit.ini` files, it was nice to be able to simplify the commands.
- Data in the module namespace is harder to verify, because you can't assume attributes in there are relevant to flit. This is alleviated by having most of it in a dedicated `__FLIT__` namespace, though.
- @Carreau has pointed out to me the advantages of being easily able to generate and modify the metadata programmatically. That's probably not impossible with it embedded in the module, but it's certainly a lot harder and more reliable than with the data in a dedicated declarative file.

As a specific point about the current implementation, I also don't like that it uses a function imported from flit inside the package: this means that every package would need a runtime dependency on flit, and that feels too invasive for a packaging tool. My aim is that people should be able to install and use flit-produced packages without flit ever being involved. So I'd rather the data was declared as a plain dictionary and validated by flit when it loads it - even though I appreciate that this loses the benefit of IDEs completing the parameters.

I recall what you said on twitter about the redundant data when packages specify `__author__` internally but still need author in the config file. I wonder if we should look at a more targeted change, where particular values could come from either the config file or the module. I'm also in two minds about that, though - I do like the principle that there should be only one way to do things, and I'm not in the habit of using `__author__` in my own packages.

I'll keep on thinking :-)

cjrh:
> On 12 Jul 2015, at 7:00 am, Thomas Kluyver <notifications@github.com> wrote:
> As a specific point about the current implementation, I also don't like that it uses a function imported from flit inside the package: this means that every package would need a runtime dependency on flit, and that feels too invasive for a packaging tool. My aim is that people should be able to install and use flit-produced packages without flit ever being involved.
Yes, I agree 100%.  This is the main problem.  My idea is not viable because of this issue.  Either you get the IDE stuff, but need to have flit installed, or, you don’t need flit to use the package, but no IDE stuff.  It’s a tricky one.

One idea for best-of-both-worlds I thought of was the following:
 flit provides a tool for package authors that creates a new file inside your package called, say, flit_config.py. This becomes part of your package.
This new file flit_config.py contains only the function for setting __FLIT__ (this is how you keep IDE support)
Then, in your module/package, you set __FLIT__ using the function imported from this new local flit_config.py
Flit itself still does all the package building as in the PR (i.e. it looks for and finds __FLIT__, proceeds as in the PR.  flit importing the new local flit_config.py might be tricky, needs more thought)
When the package is installed on another machine, the only file that is needed, and which is now part of your package anyway, is that local flit_config.py; flit itself is not required.

So what have we gained? We still get the IDE completion for setting params, and we still have the metadata inside the module/package (SSOT), and we no longer need flit to be installed on the user’s machine where the package gets installed. 

What have we lost? Well, now we have an additional file (i.e., flit_config.py) which is what we were trying to avoid in the first place. Also, this file is going to be copied over and over again for each package; granted, it’s kinda similar in that respect to setup.py so it’s not significantly worse.  Secondly, given that we would now have committed to carrying an additional file, it isn’t clear that this one is significantly better than the very simple .ini file in the first place.  

I think the current `flit init` weakens the need for any IDE completion, since you already get a template with required fields.  Is there a facility for using a pre-configured template from one’s home folder, e.g. `.flit.ini.template` to create a new one?  `flit init` would check the home folder for this file, and if it exitsts, use that to create the ini file.  This could be pre-configured for all one’s own author settings for example.
> I'm also in two minds about that, though - I do like the principle that there should be only one way to do things, 
> 
Agree 100%.  I really don’t like that metadata appears in multiple places. I think you’re better off keeping as much as possible in the .ini, as you currently.

cjrh:
I had a good day to think about it. It seems pretty clear the .ini files offer the most simple interface.

Carreau:
`flit_config.py` seem like sleepy slope to `setup.py` but there are good idea here. I definitively see ore and more that a "Package" shoudl have a way to be more aware of it's version in some places.

takluyver:
Thanks again for all these thoughts, @cjrh - I really appreciate consideration of different ways to do this.

Out of interest, are there any other fields besides `author` that you find yourself duplicating in the ini file and the package code?

I've resisted adding too much complexity to `flit init`, because I want to keep flit focussed on building packages rather than generating config files. There are tools already out there dedicated to setting up new projects, like [cookiecutter](https://github.com/audreyr/cookiecutter), and I think there's more room to experiment with building new tools for that. @Carreau convinced me that the convenience of `flit init` was worth a bit of added code, but I think doing a really polished utility like that should be a separate project.

cjrh:
@takluyver I haven't used flit before, so I can't talk much about duplication.  I have however been thinking of different ways to do what flit is doing for quite some time.  If I were using flit (I'm pretty sure I _will_ start using it for my own packages), I would try to keep as much as possible in the `.ini`.

By the way: I think that flit, or something like it, is a very much needed tool in the packaging space, to make it easier to create python packages.

I really like the cookiecutter project, but I fear that might be overkill for a better `flit init`.  I've run out of free time, but I'll think a bit more about a user template for `flit init`.  As I said, being able to set up your own custom info in a template in your home folder, and then `flit init` in a new project folder makes a new `flit.ini` with your template data seems like a neat way to get going.  This can't be much more than twenty lines of new, low-complexity code.

takluyver:
Thanks - I'm really happy to have created something people find useful, and I'll remain happy if it provokes thought that leads someone to a better solution that ultimately replaces flit.

Re tools for generating the ini file: I'm not saying it needs to be something like cookiecutter. I've thought about maybe doing a curses-style UI with the fields prefilled with guesses instead of the series of questions, or of writing the guesses straight to the file and opening it in an editor. But I think it's also an open question whether you want a fleet of small tools for 'add X to my project' (e.g. X=flit.ini), or whether you want something that sets up a project with the different pieces you want (tests, docs, packaging...) to start with. Or indeed one tool that can add individual pieces on demand.

There's a lot of interesting possibilities there, and I don't want to pre-empt exploration by welding one particular model into flit. So I'm likely to resist even relatively small improvements to flit init, to try to push people into building what they want as a separate tool.

PR #45, opened 2015-07-12
add debug messages

Carreau:


flit/wheel.py
```diff
@@ -45,11 +47,21 @@ def clean_build_dir(self):
...
+
+
         mod =  self.module
         # Copy module/package to build directory
         if mod.is_package:
             ignore = shutil.ignore_patterns('*.pyc', '__pycache__')
-            shutil.copytree(str(mod.path), str(self.build_dir / mod.name), ignore=ignore)
+            shutil.copytree(str(mod.path), str(self.build_dir / mod.name),
+                    ignore=ignore,
+                    symlinks=True,
```
takluyver: Is there any use in setting this? It's either going to fail later or lose the symlink relationships when it creates the zip file, right?

Carreau: Slipped through `add -p`.

Carreau:
fixed.

flit/wheel.py
```diff
@@ -8,6 +8,8 @@
 from . import common
 from . import inifile
 
+from shutil import copy2, copytree
```
takluyver: shutil is already imported above - for neatness, let's use `shutil.copy2` below and drop the extra import

Carreau:
done.

PR #46, opened 2015-07-12
Add 'flit develop' command

jhamrick:
This adds a `flit develop` subcommand which acts the same as `flit install --symlink` with the exception that it also installs any development requirements that are specified in the `flit.ini` file (under `develop-requires`).

I'll add tests and update relevant docs if you think this looks ok.

### DIFF
--- a/flit/inifile.py
+++ b/flit/inifile.py
@@ -20,6 +20,7 @@ class ConfigError(ValueError):
 metadata_list_fields = {
     'classifiers',
     'requires',
+    'develop-requires'
 }
 
 metadata_allowed_fields = {
@@ -93,7 +94,7 @@ def verify_classifiers(classifiers):
     _verify_classifiers_cached(classifiers)
 
 
-def read_pkg_ini(path):
+def read_pkg_ini(path, develop=False):
     """Read and check the -pkg.ini file with data about the package.
     """
     cp = configparser.ConfigParser()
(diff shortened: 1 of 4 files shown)


Carreau:
+1

Carreau:
Note from :beers: , check if/how to use `test-requirements` metadata for that.

takluyver:
I'm happy with the basic idea, but I'm still thinking a bit about what the interface should look like.

flit/inifile.py
```diff
@@ -93,7 +94,7 @@ def verify_classifiers(classifiers):
     _verify_classifiers_cached(classifiers)
 
 
-def read_pkg_ini(path):
+def read_pkg_ini(path, develop=False):
```
takluyver: I don't think the options you pass in should reach this level - the `inifile` module is to read and check the config file, and that process shouldn't be affected by what you want to use it for.

jhamrick:
Ok, updated as per our conversation in person. Got rid of `flit develop` and modified `flit install` so that it also installs any development requirements, if they are given. Additionally, I added `-s` to the `--symlink` flag so that you can do `flit install -s` to get a proper development install rather than having to type the full `flit install --symlink`.

takluyver:
Thanks, I think this looks good.

I'd quite like to have a way to control which dependencies get installed before we release again. You're welcome to work on this if you'd like to, or I'll hopefully get to it soon.

PR #47, opened 2015-07-14
Add flag to switch between sets of dependencies

jhamrick:
The options to the `--deps` flag are 'all', 'install', 'develop', and 'none'. The default is 'all', which installs both the 'install' and 'develop' dependencies. If 'none' is specified, then no dependencies are installed.


takluyver:
I think this looks nice, but I'd like a better name for 'install'. Maybe 'main'

Carreau:
the `install/main` flag is usually `production`

takluyver:
I'd be OK with calling it that

jhamrick:
Ok, updated.

Carreau:
:cake: I'm happy haven't felt such a productive day with awesome project moving forward for a long time !

jhamrick:
:confetti_ball:

Issue #48, opened 2015-08-20
flit install on self fails


dsblank:
I'm getting an error (flit installed from pip3) on flit's install with:

```bash
$ flit install
Installing requirements                                                                                                 I-flit.install
Exception:
Traceback (most recent call last):
  File "/usr/lib/python3/dist-packages/pip/basecommand.py", line 122, in main
    status = self.run(options, args)
  File "/usr/lib/python3/dist-packages/pip/commands/install.py", line 262, in run
    for req in parse_requirements(filename, finder=finder, options=options, session=session):
  File "/usr/lib/python3/dist-packages/pip/req.py", line 1632, in parse_requirements
    req = InstallRequirement.from_line(line, comes_from, prereleases=getattr(options, "pre", None))
  File "/usr/lib/python3/dist-packages/pip/req.py", line 173, in from_line
    return cls(req, comes_from, url=url, prereleases=prereleases)
  File "/usr/lib/python3/dist-packages/pip/req.py", line 71, in __init__
    req = pkg_resources.Requirement.parse(req)
  File "/usr/share/python-wheels/setuptools-3.3-py2.py3-none-any.whl/pkg_resources.py", line 2667, in parse
    reqs = list(parse_requirements(s))
  File "/usr/share/python-wheels/setuptools-3.3-py2.py3-none-any.whl/pkg_resources.py", line 2605, in parse_requirements
    line, p, specs = scan_list(VERSION,LINE_END,line,p,(1,2),"version spec")
  File "/usr/share/python-wheels/setuptools-3.3-py2.py3-none-any.whl/pkg_resources.py", line 2583, in scan_list
    "Expected ',' or end-of-list in",line,"at",line[p:]
ValueError: ("Expected ',' or end-of-list in", 'pexpect>=3.3;', 'at', ';')

Storing debug log for failure in /home/dblank/.pip/pip.log
Traceback (most recent call last):
  File "/usr/local/bin/flit", line 11, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.4/dist-packages/flit/__init__.py", line 75, in main
    Installer(args.ini_file, user=args.user, symlink=args.symlink, deps=args.deps).install()
  File "/usr/local/lib/python3.4/dist-packages/flit/install.py", line 176, in install
    self.install_requirements()
  File "/usr/local/lib/python3.4/dist-packages/flit/install.py", line 158, in install_requirements
    check_call(cmd)
[...]

PR #49, opened 2015-08-24
Trailing newline on requirements.txt

takluyver:
Fixes gh-48, I hope


takluyver:
Closing as it appears we need pip 7 anyway, and in that case, this isn't an issue.

PR #55, opened 2015-11-04
Die more gracefully on empty only-whitespace docstring.

Carreau:
`docstring_lines[0],` was crashing, technically I am a little to harsh on the check but that's for teh good cause.


takluyver:
Looks good to me!

How are things in Berkeley? I'm off to London tomorrow for Mozfest over the weekend.

PR #57, opened 2015-12-02
Log.info for script (same level as where module written)

Carreau:
See line 164-175 have log.info, so for consistency...


takluyver:
That's probably reasonable. Thanks!

PR #59, opened 2015-12-06
Add ASCII art logo

Carreau:
Because I'm in an artistic mood.


Carreau:
Looks better in a real terminal. 

screen shot 2015-12-06 at 19 22 31

takluyver:
:-D I love it! Let's put the logo import inside the `if`, so users aren't paying that tiny performance penalty every time.

If you want artistic feedback, I think the wing looks a bit oddly narrow nearest the body. Maybe try bringing the trailing edge down a bit? And is it practical to suggest the other wing behind it, or is that too fine a detail? But if you don't want to do changes, I'm happy to merge it as is :-)

It was a lovely surprise the first time I saw a hummingbird in Berkeley, so I like that they've found their way into one of my projects.

Carreau:
Made wings a bit wider and doubled it. Fixed import. 

Is that better ?

takluyver:
:+1: :cake:

PR #60, opened 2015-12-06
Context Manager to modify config.

Carreau:
Test will fail as `_validate_config`, actually pop values from config.

### DIFF
--- a/flit/inifile.py
+++ b/flit/inifile.py
@@ -102,7 +102,7 @@ def verify_classifiers(classifiers):
 
 
 def read_pkg_ini(path):
-    """Read and check the -pkg.ini file with data about the package.
+    """Read and check the `flit.ini` file with data about the package.
     """
     cp = _read_pkg_ini(path)
     return _validate_config(cp, path)
@@ -117,8 +117,7 @@ def _read_pkg_ini(path):
 
 def _validate_config(cp, path):
     """
-    validate a config section and return a strip down version fro Metadata
-
+    Validate a config and return a dict containing `module`,`metadata`,`script`,`entry_point` keys.
     """
     unknown_sections = set(cp.sections()) - {'metadata', 'scripts'}
     if unknown_sections:
--- a/flit/inifile.py
+++ b/flit/inifile.py
@@ -3,6 +3,7 @@ import difflib
 import logging
 import os
 from pathlib import Path
+from contextlib import contextmanager
 import sys
 
 import requests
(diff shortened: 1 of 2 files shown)


takluyver:
What's the use case you want this for?

If we're doing programmatic modification of config files, I'd like to look into round-tripping parsers (e.g. the [iniparse](https://pypi.python.org/pypi/iniparse) module, but preferably something that's still active). With the standard library configparser module, any formatting or comments you have in the config file is lost when you overwrite it. And comments are one of the main reasons to use ini instead of JSON.

Carreau:
> If we're doing programmatic modification of config files,

Yes, programatic modification.

Would you be ok in just splitting the validate logic out, so that I can use a rough programatic way to modify ?


We can also have that in flit (private) and have round-tripping as a later feature.

takluyver:
You mean splitting validate into a separate function from reading
configuration? Sure, I don't see a problem with that.

Carreau:
> You mean splitting validate into a separate function from reading
> configuration? Sure, I don't see a problem with that.

Done, removed the context manager and change the validate method not to modify the config object it got as parameter.

takluyver:
Thanks. At the airport now, but I'll try to have a look this evening from
the hotel.

Carreau:
Ping BTW.

flit/inifile.py
```diff
@@ -104,10 +104,22 @@ def verify_classifiers(classifiers):
...
     cp = configparser.ConfigParser()
     with path.open() as f:
         cp.read_file(f)
 
+    return cp
+
+
+def _validate_config(cp, path):
+    """
+    validate a config section and return a strip down version fro Metadata
```
takluyver: * 'a config section' -> config (it's not a section, it's the whole thing, but there isn't a good word for that)
* It's not exactly stripped down - the config format includes some extra things under `[metadata]` that aren't strictly part of the distribution metadata. And it returns a dictionary containing several things, only one of which is metadata.

flit/inifile.py
```diff
@@ -104,10 +104,22 @@ def verify_classifiers(classifiers):
 def read_pkg_ini(path):
     """Read and check the -pkg.ini file with data about the package.
```
takluyver: Nothing to do with this PR, but we should update this docstring to say flit.ini

flit/inifile.py
Carreau: How's 

```
Validate a config and return a dict containing `module`,`metadata`,`script`,`entry_point` keys
```

flit/inifile.py
Carreau: Done.

takluyver:
Thanks

PR #61, opened 2015-12-14
More extensive check os version number

Carreau:
PyPA `packaging` already normalize it, just need to check that normalization===identity.

### DIFF
--- a/flit/common.py
+++ b/flit/common.py
@@ -9,13 +9,14 @@ log = logging.getLogger(__name__)
 import re
 
 
-pep440re = re.compile('^([1-9]\d*!)?'        # [N!]
-                      '([1-9]\d*)'           # N
-                      '(.[1-9]\d*)*'         # (.N)*
-                      '((a|b|rc)[0-9]\d*)?' # [{a|b|rc}N]
-                      '(\.post[1-9]\d*)?'         # [.postN]
-                      '(\.dev[1-9]\d*)?$'         # [.devN]
-                      )
+pep440re = re.compile('^'
+                   '([1-9]\d*!)?'
+                 '(0|[1-9]\d*)'
+               '(.(0|[1-9]\d*))*'
+        '((a|b|rc)(0|[1-9]\d*))?'
+          '(\.post(0|[1-9]\d*))?'
+           '(\.dev(0|[1-9]\d*))?'
+        '$')
 
 def _is_canonical(version)->bool:
     """
--- a/flit/common.py
+++ b/flit/common.py
@@ -19,7 +19,7 @@ pep440re = re.compile('^([1-9]\d*!)?'        # [N!]
 
 def _is_canonical(version)->bool:
     """
-    Retunr wether or not the version string is canonical according to Pep 440
+    Return whether or not the version string is canonical according to Pep 440
     """
     return pep440re.match(version) is not None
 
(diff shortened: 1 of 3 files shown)


takluyver:
I'd rather avoid depending on packaging so long as it's practical, and in this case I think it is. PEP 440 specifies this format:

```
[N!]N(.N)*[{a|b|rc}N][.postN][.devN]
```

So we just need to translate that to a regex and check that it matches `__version__`. I might warn rather than throwing an error, as versions that don't match that are still allowed.

Carreau:
> So we just need to translate that to a regex and check that it matches __version_

[There](https://github.com/Carreau/pep440/blob/master/pep440/__init__.py#L25-L31https://github.com/Carreau/pep440/blob/master/pep440/__init__.py#L25-L31)

I did it this afternoon. It might be a thing to add into the pep itself maybe.

Carreau:
In case you are lazy:

```
pep440re = re.compile('^([1-9]\d*!)?'        # [N!]
                      '([1-9]\d*)'           # N
                      '(.[1-9]\d*)*'         # (.N)*
                      '((a|b|rc)[0-9]\d*)?' # [{a|b|rc}N]
                      '(\.post[1-9]\d*)?'         # [.postN]
                      '(\.dev[1-9]\d*)?$'         # [.devN]
                      )
```

takluyver:
Nice. Spelling PR coming up ;-)

Carreau:
> Nice. Spelling PR coming up ;-)

Feel free to push/release on PyPI.

Carreau:
Removed dependency. Still raising. Do you prefer log-info , log-warn, log-debug, or keep raise ?

flit/common.py
```diff
@@ -62,19 +80,32 @@ def get_info_from_module(target):
...
+    if not isinstance(version, str):
         raise InvalidVersion('__version__ must be a string, not {}.'
-                                .format(type(module_version)))
-    if not module_version[0].isdigit():
+                                .format(type(version)))
+    if not version[0].isdigit():
         raise InvalidVersion('__version__ must start with a number. It is {!r}.'
-                                .format(module_version))
+                                .format(version))
+    if _is_canonical(version):
```
takluyver: I think you're missing a `not` here ;-)

flit/common.py
```diff
@@ -6,6 +6,24 @@
...
+                      '([1-9]\d*)'           # N
+                      '(.[1-9]\d*)*'         # (.N)*
+                      '((a|b|rc)[0-9]\d*)?' # [{a|b|rc}N]
+                      '(\.post[1-9]\d*)?'         # [.postN]
+                      '(\.dev[1-9]\d*)?$'         # [.devN]
+                      )
+
+def _is_canonical(version)->bool:
+    """
+    Retunr wether or not the version string is canonical according to Pep 440
```
takluyver: 'Return whether'

takluyver:
I think a warning would be the appropriate level - it indicates the user is doing something non-optimal but still allowed.

flit/common.py
Carreau: `not` not not added.

flit/common.py
Carreau: Fixed.

Carreau:
And warning.

Carreau:
Need to fix docstring/test.

Carreau:
I think I need to rethink the regex, as things can be `0|[1-9]\d+` like `dev0` is allowed.

Carreau:
OK, works now.

takluyver:
:+1:

PR #63, opened 2015-12-27
[Idea] Parse the AST, to guess minimal Python version.

Carreau:
Small draft prototype with `yield from`, which obviously is not **that** useful as it is 3.3+ only, but already eliminate Python2. 

What do you think of a feature like that ? I think that would push the usage of the `requires-python` field in metadata.

(Merry Christmas also ! )

### DIFF
--- a/flit/__init__.py
+++ b/flit/__init__.py
@@ -48,7 +48,7 @@ def main(argv=None):
     parser_install.add_argument('--deps', choices=['all', 'production', 'develop', 'none'], default='all',
         help="Which set of dependencies to install")
 
-    parser_init = subparsers.add_parser('init',
+    subparsers.add_parser('init',
         help="Prepare flit.ini for a new package"
     )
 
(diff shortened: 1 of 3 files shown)


takluyver:
Merry Christmas! Or indeed joyeux Noël.

I think I'm going to apply a principle we use sometimes in IPython here: if you can't make a feature work well and reliably, it may be better not to provide it at all.

Why can't this be made to work reliably? First, there are many changes beyond new syntax. Code using `subprocess.run()`, for example, requires Python 3.5. Detecting this is possible, but gets much more complex, especially with backport modules involved. Secondly, it's possible to have one module in a package require a newer version than the package as a whole - e.g. in Pexpect, `pexpect.async` uses `yield from`, but Pexpect overall can be installed and used on Python 2.7 (just not that module).

As you know, I also don't want to add many features to `flit init`, but I'm happy for separate tools to try doing more complicated things like this. ;-)

Thanks, though, it is an interesting idea.

PR #64, opened 2016-01-04
Reproducible builds

takluyver:
@Carreau :

```
[takluyver@tk2e15 flit]$ flit wheel
[takluyver@tk2e15 flit]$ md5sum dist/flit-0.7.2-py3-none-any.whl 
104fa2df7ba06b4c1369299761dc9866  dist/flit-0.7.2-py3-none-any.whl
[takluyver@tk2e15 flit]$ flit wheel
[takluyver@tk2e15 flit]$ md5sum dist/flit-0.7.2-py3-none-any.whl 
104fa2df7ba06b4c1369299761dc9866  dist/flit-0.7.2-py3-none-any.whl
```

There's two levels to this. If you run `flit wheel` twice on the same source tree locally, you should get the same results. But timestamps in the zip file reflect the time when you last pulled changes from git (git doesn't store mtimes), so someone else can't immediately reproduce the build from another clone. To make a properly reproducible build, set [SOURCE_DATE_EPOCH](https://reproducible-builds.org/specs/source-date-epoch/). Try this (on this branch) and see if you get the same result:

```
[takluyver@tk2e15 flit]$ export SOURCE_DATE_EPOCH=1451922369
[takluyver@tk2e15 flit]$ flit wheel
Zip timestamps will be from SOURCE_DATE_EPOCH: 2016-01-04 15:46:09  I-flit.wheel
[takluyver@tk2e15 flit]$ md5sum dist/flit-0.7.2-py3-none-any.whl 
8a49f357088a0596810566989a428b82  dist/flit-0.7.2-py3-none-any.whl
```

### DIFF
--- a/flit/wheel.py
+++ b/flit/wheel.py
@@ -1,5 +1,6 @@
 import configparser
 import contextlib
+from datetime import datetime
 import hashlib
 import io
 import logging
@@ -45,6 +46,19 @@ class WheelBuilder:
 
         self.dist_version = self.metadata.name + '-' + self.metadata.version
         self.records = []
+        try:
+            # If SOURCE_DATE_EPOCH is set (e.g. by Debian), it's used for
+            # timestamps inside the zip file.
+            d = datetime.utcfromtimestamp(int(os.environ['SOURCE_DATE_EPOCH']))
+            log.info("Zip timestamps will be from SOURCE_DATE_EPOCH: %s", d)
+            # zipfile expects a 6-tuple, not a datetime object
+            self.source_time_stamp = (d.year, d.minute, d.day, d.hour, d.minute, d.second)
+        except (KeyError, ValueError):
+            # Otherwise, we'll use the mtime of files, and generated files will
+            # default to 2016-1-1 00:00:00
+            self.source_time_stamp = None
+
+        # Open the zip file ready to write
         self.wheel_zip = zipfile.ZipFile(str(self.wheel_path), 'w',
                              compression=zipfile.ZIP_DEFLATED)
 
@@ -65,7 +79,8 @@ class WheelBuilder:
     def _add_file(self, full_path, rel_path):
         log.debug("Adding %s to zip file", full_path)
         full_path, rel_path = str(full_path), str(rel_path)
-        hashsum = my_zip_write(self.wheel_zip, full_path, rel_path)
+        hashsum = my_zip_write(self.wheel_zip, full_path, rel_path,
+                               date_time=self.source_time_stamp)
         size = os.stat(full_path).st_size
         self.records.append((rel_path, hashsum.hexdigest(), size))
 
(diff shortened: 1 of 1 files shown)


takluyver:
For some reason we're creating an invalid zip file under Python 3.5, though it's fine under 3.4. I'm looking into what's changed in the zipfile module.

Carreau:
Sweet ! (just drove 6h almost non stop, but will look at that later)

takluyver:
OK, got the zipfile stuff working with Python 3.5. Of course that means that the md5sum of a build from this branch has changed:

```
[takluyver@tk2e15 flit]$ echo $SOURCE_DATE_EPOCH 
1451922369
[takluyver@tk2e15 flit]$ flit wheel
Zip timestamps will be from SOURCE_DATE_EPOCH: 2016-01-04 15:46:09  I-flit.wheel
[takluyver@tk2e15 flit]$ md5sum dist/flit-0.7.2-py3-none-any.whl 
472b266d8efa0724da385797fe0228f7  dist/flit-0.7.2-py3-none-any.whl
```

Carreau:
Awesome !

What do you think (independently) of having a ZipFile that can built in-memory zips?

Carreau:
> What do you think (independently) of having a ZipFile that can built in-memory zips?

Precision: not the final zip file in memory, but something that make what we are doing here easier. Basicaly having [ZipFile.write](https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile.write) take a "Fake file" with `os.stats` info and so on... instead of a file name ?

takluyver:
I'm working on a patch for zipfile to add more flexibility along those
lines. Currently, it's a new method that takes a file like object and a
ZipInfo object.

Carreau:
Will be happy to review it. (and looking forward CPython being on GitHub)

takluyver:
Python issue for making zipfile more flexible: http://bugs.python.org/issue26039

I've included a patch, but it's more a starting point to work out the API than a definite proposal.

takluyver:
Ping so you don't forget to have a look at this, @Carreau :wink:

I'm sure it's a good way to hold sleep off as you adjust time zones.

flit/wheel.py
```diff
@@ -156,12 +163,120 @@ def post_build(self):
...
+
+        self.wheel_zip.close()
+
         self.post_build()
+
+
+import stat, time
+from zipfile import ZipInfo, ZIP_LZMA, _get_compressor, ZIP64_LIMIT, crc32
+
+def my_zip_write(self, filename, arcname=None, compress_type=None,
```
Carreau: I would just mark the line/section you changed in here, of add a comment with version of Python you stole that from. 
(easier for later update).

+1 otherwise. 

BTW. Flit made it to pycoders weekly [top projects of 2015](http://us4.campaign-archive2.com/?u=9735795484d2e4c204da82a29&id=22fde1c28d&e=db2b52b812)

takluyver: It's actually now a bit of a hodgepodge of code from Python 3.4 and 3.5, because neither version would quite work with the other. Hopefully by 3.6 we can add functionality to `zipfile` so this is no longer necessary.

Nice! I noticed I'd got another ~20 stars. Flit is slowly closing on Pynsist as my most starred project.

takluyver:
I was able to produce bit-for-bit identical wheels on different machines using this branch and setting `$SOURCE_DATE_EPOCH`. However, as it stores file permissions, it is sensitive to the umask on the machine where you checked out the repository - this initially caused me some trouble, because Debian and Fedora have different default umasks. In the future, I may look into normalising permission bits, but I think it's good enough for now.

@Carreau I'll merge this soon unless you want to look at it any more.

Carreau:
Go ahead and merge. I'm scraping craiglist for a place to stay most of the day, so am already behind schedule on too  many things.

takluyver:
Good luck. You should approach it as a data science problem - use machine learning to pick where you're going to live. Live in many places for a few days each and rate them to train the model.

PR #65, opened 2016-01-05
Add installfrom subcommand

takluyver:
This adds a way to install directly from a source tarball or a Github repository.

### DIFF
--- a/flit/__init__.py
+++ b/flit/__init__.py
@@ -11,6 +11,13 @@ __version__ = '0.7.2'
 
 log = logging.getLogger(__name__)
 
+def add_shared_install_options(parser):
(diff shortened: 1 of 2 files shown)


Carreau:
why `installfrom` and not use this if argument of install start with `https/ssh/....` ?

takluyver:
I thought about that too. But at present, `install` doesn't take an argument, it just works on the current directory, so it didn't seem quite right to give it an optional argument. Maybe that is preferable, though.

takluyver:
I've made sure that it works with a Github repository, and having thought about it a bit more, I still prefer this interface to overloading the `install` command. So, landing lights on...

PR #66, opened 2016-01-17
Normalise permission bits to 755 or 644

takluyver:
Apparently, the only permission bit git tracks is executable or not. So this flattens the permission bits to 755 for executable files, or 644 for non-executable, matching the default umask on Debian systems.

I still want to think about this:
- Does wheel installation preserve any more permission bits? If not,
  this is definitely fine.
- Do other version control systems store more permissions?
- Is there a reason you might want other permissions on files installed
  from wheels, e.g. to make a file not world readable? It seems redundant
  when packages come from a public index, but maybe for privately
  installed packages it makes sense.


Carreau:
> Apparently, the only permission bit git tracks is executable or not

What about other versioning system. I though you didn't want to special case git in flit ?

takluyver:
I don't. That's why it's on my list of things to check. ;-)

brettcannon:
I just checked Mercurial and Subversion and both don't track more than execution.

takluyver:
Thanks Brett for investigating that. With that and the existence of `distutils644`, which I heard about recently, I think this is the right thing to do.

PR #68, opened 2016-01-21
Fix windows incompatibility

blink1073:
I was able to use flit on Windows after making this change.

https://docs.python.org/3/library/os.html#os.getuid


takluyver:
Thanks, good catch

PR #69, opened 2016-01-23
Allow extra config sections, and pass through raw config for extensions

takluyver:
I'm thinking about how to implement flonda (flit to conda), and I want to use an extra section in the flit.ini file. Currently, this would make flit throw an error. This turns that error into an info-level log message, and passes the config object through so extensions can extract extra information.

An alternative is to require some prefix for extensions, e.g. `[x-flonda]`, similar to X- HTTP headers.


takluyver:
Merged #70 instead

PR #70, opened 2016-01-27
Allow extra config sections with x- prefix

takluyver:
Alternative to gh-69. I'm leaning towards this one, but not by much.


takluyver:
I'm going for this one. It can always be made more flexible later.

:shipit:

PR #72, opened 2016-02-02
Install with pip

takluyver:
With this change, `flit install` (without `--symlink`) builds a wheel and invokes `pip` to install it. This should resolve issues @jhamrick saw previously where a package was installed to the wrong target directory. It should also result in `.exe` wrappers on Windows, rather than `.bat` wrappers.

`flit install` and `flit installfrom` also gain a `--python` parameter to run on another Python installation. E.g.:

```
flit install --python python2.7   # Legacy Python!
flit install --python /path/to/an/env/bin/python
source activate another_env
flit install --python python  # This targets the Python in the env
```

The default is still the Python which flit is running on. Flit doesn't need to be installed for the target Python (so it works on Python 2).

### DIFF
--- a/flit/__init__.py
+++ b/flit/__init__.py
@@ -18,6 +18,9 @@ def add_shared_install_options(parser):
     parser.add_argument('--env', action='store_false', dest='user',
         help="Install into sys.prefix (default if site.ENABLE_USER_SITE is False, i.e. in virtualenvs)"
     )
+    parser.add_argument('--python', default=sys.executable,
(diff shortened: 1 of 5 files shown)


Carreau:
Sorry for the lack of review, busy interviewing to hire ~~your substitute~~ more post docs.

takluyver:
No problem, I know finding someone to take my place is going to be difficult ;-)

Carreau:
> No problem, I know finding someone to take my place is going to be difficult ;-)

You are unique !

PR #77, opened 2016-03-27
Implement PEP 517, store data in pyproject.toml

takluyver:
I'm still not sure about this, but this is roughly what it takes to add TOML support for config.
- TOML is more verbose than INI, but better specified and more flexible. Explicit lists are nicer than multi-line strings (IMO), and it makes it easier to have entrypoints defined in the main config file rather than in a separate file (though we could do that in ini too).
- TOML is younger, so there's less editor support - PyCharm doesn't have syntax highlighting for TOML, while it does for ini.
- There's some extra complexity to support two different config formats. This can't go away any time soon, because dropping ini support would break existing projects using flit.
- It adds a dependency by someone I don't know ([uiri](https://github.com/uiri)'s `toml` package). Recent events in the JS ecosystem have highlighted the risks of adding lots of external dependencies ([is-positive-int was](https://github.com/tjmehta/is-positive-integer/pull/2/files) a remarkable example).

I think I'm currently leaning towards being conservative and sticking with ini for now. I like toml, but it's not clear that switching is a net benefit for this use case.

### DIFF
--- a/flit.ini
+++ b/flit.ini
@@ -16,3 +16,4 @@ classifiers=Intended Audience :: Developers
 
 [scripts]
 flit = flit:main
+


Carreau:
do you want to update that to use pyproject.toml ?

takluyver:
I'm waiting for the other part of the build system abstraction stuff, that tells tools how to run the build system. Once that's done, I'll look at using pypackage.toml.

dholth:
I currently prefer pytoml, it seems to be well tested. toml has some bugs: `toml.loads(toml.dumps({'foo=bar': ['sdl2_lib']}))`

takluyver:
Thanks for the info. It looks like pytoml is also more actively maintained. I'll probably switch when I come back to this.

Do you know if there's any movement on PEP 516/517 to follow on from PEP 518?

dholth:
Haven't heard anything.

takluyver:
I've started updating this to PEP 518's `pyproject.toml`. I'm still [trying](https://github.com/python/peps/pull/51) to push the other relevant PEPs forwards.

There's not now a clear difference in maintenance between toml and pytoml, and the bug @dholth mentioned in toml seems to be fixed. So I'm sticking with toml for the time being, but I'm open to reviewing that if one looks better than the other.

dholth:
toml is nice because it is just one file.  glad to hear they fixed the bug.

takluyver:
This now provides an implementation of the buildsystem API provisionally specified by [PEP 517](https://www.python.org/dev/peps/pep-0517/).

I'm still thinking about the best way to specify dependencies in the TOML config.

PR #87, opened 2016-09-04
Support "Compatible With" syntax for Legacy Python detection.

Carreau:
`~=X.y` meaning `>=X.y` and `<(X+1)`.


takluyver:
Thanks

PR #89, opened 2016-09-17
Provide command to generate a setup.py file.

Carreau:
This is useful for downstream distribution to generate a setup.py when
they cannot rely on wheels to install packages; this is a stopgap while
pep 517 is not implemented.

---

Wrote that as i was requested to add a setup.py for [a package of mine](https://github.com/Carreau/nose_warnings_filters/issues/4), I know it can be controversial, so I'll understand if you say no.

### DIFF
--- a/flit/__init__.py
+++ b/flit/__init__.py
@@ -61,8 +61,12 @@ def main(argv=None):
     )
     add_shared_install_options(parser_installfrom)
 
-    parser_init = subparsers.add_parser('init',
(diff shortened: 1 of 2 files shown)


Carreau:
(Todo entry points)

takluyver:
Hmm. I'll consider it, but I'm not enthusiastic, for the reasons already stated.

For the general case, this is also missing `package_data`, though there are many packages that could work without that.

PR #90, opened 2016-09-18
Test on Nightly but allow_failure.

Carreau:


takluyver:
I'm not sure if it's useful to test but allow failures - I only go and look at Travis when there is a failure, so I won't usually see any failures that only occur on nightly. Can Travis build with the 3.6 betas yet? If so,  I might add that as a target and not use `allow_failure`.

Carreau:
> Can Travis build with the 3.6 betas yet? If so, I might add that as a target and not use allow_failure.

Not yet (AFAICT) and it seem to me that  current "nightly" point to the 3.6 branch.

Agreed that once we have a 3.6 target we can add it without allow_failures. The problem for now is that most of the failures are due to bug upstream so almost all of repo would fail with unfixable errors. While ipykernel pass (which I'm surprised about) everything that touch traitlets will be kill because of an out of memory error.

Testing on nightly ( or for that matter 3.6-betas once that target exists) allows us to know when the upstream bugs are fixed.

Travis Build of nightly seem also a bit buggy:
```
$ export PYTHONWARNINGS='ignore'; python -c 'print(1)'
'import warnings' failed; traceback:
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/warnings.py", line 496, in <module>
    _processoptions(sys.warnoptions)
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/warnings.py", line 178, in _processoptions
    _setoption(arg)
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/warnings.py", line 184, in _setoption
    import re
  File "/home/travis/virtualenv/python3.6-dev/lib/python3.6/re.py", line 122, in <module>
    import enum
ModuleNotFoundError: No module named 'enum'
```

> I only go and look at Travis when there is a failure, so I won't usually see any failures that only occur on nightly.


I check Travis more often, and i think it is useful to know when upstream have broken something, it allows you to report the bug immediately. If we had been testing on nightly beta1 would never have been released with http://bugs.python.org/issue28147 and I got another segfault corrected upstream because of nightly +allow_failures testing.

I'm also poking at travis API and looking as whether I can get all the project that test on nightly & 3.5 through time and look at the evolution of nightly-failure/3.5-success and see if I can autodetect nightly regression at a day or so granularity; Or even during the Python 3.6 Beta get an indicator of projects "Compatibility" with 3.6.

takluyver:
I don't want to do this in flit with 'nightly' at the moment:

- If we're bothering to run tests, I don't want allow_failures - I want to know about it when tests fail
- If the tests do fail, it wouldn't be obvious whether it's because of a change in flit, a change in Python, or a bug in Travis' build of Python (as in your example above).
- This is exacerbated because flit is not changing very quickly at the moment, so nightly could change a lot between flit builds, and we're unlikely to pick anything up promptly.

During the 3.5 betas, you could ask Travis to build on e.g. `3.5.0b1`. This was based on pyenv, which has recently added 3.6.0b1 (yyuu/pyenv#699). If that works on Travis, I'd be happy to start testing with the numbered betas.

PR #94, opened 2016-10-17
Skip registering of packages when response is 410.

Carreau:
Warehouse refuse registration, you now need to upload directly.


takluyver:
It can reach this in two ways:

- User explicitly ran `flit register`
- User tried to upload to a project which they don't have permission to upload

In both situations, I'd rather it failed with an error rather than proceeding. I've been thinking about checking if the URL ends with `legacy` to decide if we're targeting Warehouse. This would also be useful for the 'Package is at ...' message after a successful upload.

PR #97, opened 2016-12-28
Use entrypoints to list extra licenses.

Carreau:
Should allow some packages to expose custom licenses if they like.

--- 

Thoughts ?

### DIFF
--- a/flit.ini
+++ b/flit.ini
@@ -7,6 +7,7 @@ requires=requests
     docutils
     requests_download
     zipfile36; python_version in "3.3 3.4 3.5"
+    entrypoints
 requires-python= >=3
 description-file=README.rst
 classifiers=Intended Audience :: Developers
--- a/flit/init.py
+++ b/flit/init.py
@@ -57,6 +57,62 @@ license_names_to_classifiers = {
 
 license_templates_dir = Path(__file__).parent / 'license_templates'
 
+
+class LicenseFinder:
+
+    def __init__(self):
+        self._uptodate = False
+        self._templates = {}
+
+    def _update(self, force=False):
+        """
+        Update the know list of licences and classifiers
+        """
+        if self._uptodate and not force:
+            return
+        from entrypoints import get_group_all
+        entrypoints = get_group_all('flitlicenses')
+        for entrypoint in entrypoints:
+            name, classifier, summary, template = entrypoint.load()
+            license_choices.insert(-1, (name, summary))
+            license_names_to_classifiers[name] = classifier
+            self._templates[name] = template
+        self._uptodate = True
+
+    @property
+    def choices(self):
+        """
+        list of (name, summary) tupple
+        """
+        self._update()
+        return license_choices
+
+    @property
+    def classifiers(self):
+        """
+        Mapping from licence name to PyPI classifiers
+        """
+        self._update()
+        return license_names_to_classifiers
+
+    def get(self, name):
+        """
+        Get a license template from its name
+
+        """
+        self._update()
+        if name in license_names_to_classifiers:
+            tpl = self._templates.get(name)
+            if tpl:
+                return tpl
+            with (license_templates_dir / name).open() as f:
+                return f.read()
+        else:
+            raise ValueError("Unknown licence type: %r. Did you mean on of %r" % (
+                name, license_choices))
+
+licensesfinder = LicenseFinder()
+
 class IniterBase:
     def __init__(self, directory='.'):
         self.directory = Path(directory)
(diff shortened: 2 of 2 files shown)


takluyver:
I'm not particularly keen for flit to depend on entrypoints, because that's
packaged with flit, so it's a circular dependency for downstream packagers.
I switched zipfile36 back to setup.py before using it in flit.

Secondly, I don't really want 'flit init' to gain much more complexity. I
would like a better tool to create packages and add features, but I don't
think it should be part of flit. "Do one thing and do it well." I have some
vague ideas about this, which I'd be happy to talk about when I'm not on my
phone keyboard.

Carreau:
> I'm not particularly keen for flit to depend on entrypoints, because that's
> packaged with flit, so it's a circular dependency for downstream packagers.
> I switched zipfile36 back to setup.py before using it in flit.

Fair enough. 



> Secondly, I don't really want 'flit init' to gain much more complexity. I
> would like a better tool to create packages and add features, but I don't
> think it should be part of flit. "Do one thing and do it well." I have some
> vague ideas about this, which I'd be happy to talk about when I'm not on my
> phone keyboard.

Ok, I'll wait, but I'd really like to have choices beyond the one provided by default with flit. 
Unless we add a `5) More` option in init which allows more choice. 

Enjoy your end of 2016 and don't work too much.

dholth:
I guess you are familiar with https://spdx.org/spdx-specification-21-web-version#h.jxpfx0ykyb60

takluyver:
@dholth I'm not familiar with that. It looks interesting, but it doesn't really fit with my goal of keeping things simple here. (And is it just me who finds that document quite hard to look at?)

@Carreau OK, now I'm back at a real keyboard, so here goes.

I've been thinking about tools like cookiecutter and love, and why I don't use them, even though I've tried them and think they're neat. I reckon my complaint is that either they make a cookie that's too small, and I still have to add stuff myself, or it's too big and does things I don't want. Not every tiny package I make will have docs on RTD. Maybe this time I want to try putting it on Gitlab rather than Github, or using Mkdocs instead of Sphinx. There are many steps that I've done over and over, but I don't set up *exactly* the same structure each time, and I don't know exactly what will change. And even if I had a fixed cookie shape I used every time, other people probably wouldn't like it.

What tools like `flit init` bring is automation of adding a feature to an existing project - in this case, flit packaging. `sphinx-quickstart` is another example. Whereas cookiecutter & love expect to start with an empty directory and create everything, these tools start with a directory already containing files, and add a particular piece to it.

`flit init` and `sphinx-quickstart` are both tools that I *do* use. Sphinx quickstart annoys me a bit because it asks a lot of questions I'm not usually interested in:

```
> Name prefix for templates and static dir [_]: 
```

I have deliberately made `flit init` do the bare minimum to make a viable package, and try to be smarter about guessing defaults. But I know how to add information later in the config file, which a newcomer may not. Both are very simple `input()` based UIs, and while much more is possible if you use something like prompt_toolkit, it's a sideline, not the main point of the project, and I'm not interested in making a big thing of it. I'm guessing Sphinx isn't either, but that's a total guess.

So, I'm envisaging something different from both. Let's call it *block* for now, for the sake of description. In a code directory, you run something like `block flit`. A nice terminal UI built on prompt_toolkit appears with some guessed values inserted as defaults. You can easily select any field and replace the contents. License may be a dropdown. Perhaps there's an 'advanced' section for optional information. When you're done, you press F12 (for the sake of argument), and it creates/modifies the files as necessary.

However, `block` itself doesn't know about Flit or Sphinx. When you run `block flit`, it uses entrypoints (or a similar mechanism) to find the piece that provides that block - that could be flit itself, or a third-party package. There could be multiple for one tool, e.g. `flit`, `flit-advanced`, `flit-Carreaus-opinion`.

*Block* would provide an API for:

- Defining a form-like interface. Ideally this should be possible with minimal knowledge of prompt_toolkit, but the full power would be available to blocks that want to do cleverer things.
- Guessing default values from the working directory and from stored data.
- Storing data to inform future guesses (like `flit init` does)

This is all imagination at the moment, but I think something like that would allow us to build a more polished set of tools like `flit init` (which I'm quite fond of, even if I deny most improvements to it), without duplicating a lot of code between projects. And it could make such tools more discoverable.

Carreau:
I hear you, the issue is that now flit is doing 1/2 its job of letting users choose a licence. 
I have to do a second step if I want a BSD, while other licenses are just pressing 4 times enter. 

Anyway I can keep this patch on top of flit locally that's not an issue. 

Thanks.

PR #99, opened 2017-01-12
include license file in wheels

minrk:
metadata 2.0 specifies an extension for where to put this info, but it should work okay this way while flit is implementing 1.2.

### DIFF
--- a/flit.ini
+++ b/flit.ini
@@ -3,6 +3,7 @@ module=flit
 author=Thomas Kluyver
 author-email=thomas@kluyver.me.uk
 home-page=https://github.com/takluyver/flit
+license=LICENSE
 requires=requests
     docutils
     requests_download
--- a/flit.ini
+++ b/flit.ini
@@ -3,7 +3,6 @@ module=flit
 author=Thomas Kluyver
 author-email=thomas@kluyver.me.uk
 home-page=https://github.com/takluyver/flit
-license=LICENSE
 requires=requests
     docutils
     requests_download
--- a/flit/wheel.py
+++ b/flit/wheel.py
@@ -157,8 +157,8 @@ class WheelBuilder:
             self._add_file(self.ini_info['entry_points_file'],
                            dist_info + '/entry_points.txt')
         for base in ('COPYING', 'LICENSE'):
-            for fname in sorted(self.directory.glob(base + '*')):
-                self._add_file(fname, '%s/%s' % (dist_info, fname))
+            for path in sorted(self.directory.glob(base + '*')):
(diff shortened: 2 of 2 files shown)


flit.ini
```diff
@@ -3,6 +3,7 @@ module=flit
 author=Thomas Kluyver
 author-email=thomas@kluyver.me.uk
 home-page=https://github.com/takluyver/flit
+license=LICENSE
```
takluyver: The license field is intended for the name or description of a license ([see metadata 1.2](https://www.python.org/dev/peps/pep-0345/#license-optional)), not a path.

I think I'd rather this was a convention-over-configuration thing, looking for a file called `LICENSE` adjacent to the module.

minrk: Sure. Updated to grab `LICENSE*` or `COPYING*` to follow existing conventions.

flit/wheel.py
```diff
@@ -155,6 +156,9 @@ def write_metadata(self):
         elif self.ini_info['entry_points_file'] is not None:
             self._add_file(self.ini_info['entry_points_file'],
                            dist_info + '/entry_points.txt')
+        for name in ('LICENSE', 'COPYING'):
+            for fname in glob.glob(name + '*'):
```
takluyver: Can we sort the globbed files? For reproducible builds, we want to ensure files are added in a predictable order, and I don't think there's any guarantee about the return order from glob().

Also, this currently takes files from the working directory, which is usually but not always correct. `self.directory` is a Path object, so you should be able to use its [glob method](https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob).

minrk: Good call, done.

flit/wheel.py
```diff
@@ -155,6 +156,9 @@ def write_metadata(self):
         elif self.ini_info['entry_points_file'] is not None:
             self._add_file(self.ini_info['entry_points_file'],
                            dist_info + '/entry_points.txt')
+        for base in ('COPYING', 'LICENSE'):
+            for fname in sorted(self.directory.glob(base + '*')):
+                self._add_file(fname, '%s/%s' % (dist_info, fname))
```
takluyver: If the directory is not the cwd, its path will be added between dist_info and the filename. We can avoid this by taking `fname.name` for the second formatting argument (that's equivalent to `os.path.basename()`).

minrk: Indeed. Sorry, I'm still not used to Path objects.

takluyver: No problem, thanks for your patience.

PR #105, opened 2017-04-15
Fix misuse of flags in re.sub() calls

jwilk:
The 4th argument of `re.sub()` is maximum number of substitutions, not flags.

Found using [pydiatra](https://github.com/jwilk/pydiatra).


takluyver:
Thanks!

PR #106, opened 2017-05-11
Make sdist

takluyver:
Following discussion on #74, this code can make an sdist, including an autogenerated `setup.py`. I have tested making an sdist of flit itself and installing it using pip, confirming that it works.

To make an sdist, run `flit sdist` in the project directory. There is not yet code for uploading the sdist directly, but you can use twine to do so.

* Selecting the files to go into an sdist requires that the project is in a VCS, and that that VCS is available on PATH. `git` and `hg` are the currently supported options.
* It also checks whether there are any untracked files in the repo, or files deleted since the last commit, other than a few whitelisted ones (`dist/`, `__pycache__`, `*.pyc`). It will refuse to build the sdist if files have been added or removed which the VCS doesn't know about, but staging the changes (`git add`/`git rm`) is sufficient to make it happy. This check might be relaxed in the future.
* There is no work as yet to handle git submodules or hg subrepos, and I wouldn't expect this to correctly package projects using them.
* If there is a `setup.py` adjacent to `flit.ini` and checked into the repository, the sdist will use that instead of autogenerating a `setup.py` script. In that case, it's up to the package author to ensure that their `setup.py` is correct.
* A `pyproject.toml` is not generated; my plan is to fold `flit.ini` into that file, so it will exist in the source repo (#77).


flit/sdist.py
```diff
@@ -0,0 +1,251 @@
...
+            install_reqs.append(name_version)
+        else:
+            extra_reqs[':'+env_mark].append(name_version)
+
+    return install_reqs, dict(extra_reqs)
+
+def include_path(p):
+    return not (p.startswith('dist/')
+                or ('/__pycache__' in p)
+                or p.endswith('.pyc'))
```
Carreau: Will `/` be an issue on windows ?

(sorry for delayed review)

takluyver: Good point,  thanks. Fixed in master.

Issue #108, opened 2017-05-22
This is a great package


choldgraf:
Just FYI - super useful. Thanks :-)

feel free to close the issue ;-)

takluyver:
Thanks Chris! Has @Carreau been evangelising it to you?

It's just gaining the ability to make sdists as well, so it should be even more useful in the future.

choldgraf:
haha he did mention it, but only because I explicitly asked him if there was a package designed to make pip-ifying your module more easy...turns out there was such a package :-)

dholth:
Can reproduce.

takluyver:
:-)

Let me know if you work out a fix!

dholth:
You know it. https://bitbucket.org/dholth/enscons might make pip-ifying your packages more hard.

takluyver:
"We choose to do this, not because it is easy, but because it is hard" :-P

Part of my aim was to separate the easy cases from the hard ones. Distutils and setuptools have a lot of flexibility if you can work out how to subclass their commands, but 95% of the time we're using them to package up a few straightforward .py files. I figure that if flit makes it really easy to package straightforward pure-Python libraries, that frees other tools like enscons to focus on things like compilation and code generation.

I'm sure that eventually someone will ask for flit to support some kind of configurable build step, though.

dholth:
I like flit and mine has stalled. Thanks for keeping at it.

takluyver:
While you're both being nice about flit, could I ask for your input on something?

What commands would you look for to:
1. Create a wheel and an sdist together.
2. Create and upload a wheel and an sdist.

I've currently called these `flit build` and `flit upload`, but I'm not sure about `flit build` in particular. 'Build' might be a bit of a jargon term, and it could be confusing if we later want to add the option for a build step before making a wheel (not saying we will, but who knows). Maybe `flit package` would be better? Or maybe you have a better idea?

choldgraf:
I actually thought `flit package` when I first read point 1, then I saw you mentioned it below. I'm +1 on that.

instead of `flit upload` why not `flit push`? Fits within the broader ecosystem of "things that push'

takluyver:
Interesting idea! I might be a bit concerned about people confusing `flit push` and `git push`, though.

takluyver:
I have also wondered about `flit release` instead of `flit upload`, but I'm not convinced about that - uploading is only one step of making a release.

choldgraf:
[...]

PR #109, opened 2017-05-23
Subcommands 'flit build' and 'flit publish'

takluyver:
These allow producing and uploading both wheels and sdists at the same time.

- [x] I still want to think about the naming.
- [x] Add a way to skip one format, maybe something like `flit upload --no wheel`? Should it be positive or negative?

### DIFF
--- a/flit/__init__.py
+++ b/flit/__init__.py
@@ -47,6 +47,10 @@ def main(argv=None):
          help="Build a source distribution (.tar.gz)",
     )
 
+    parser_build = subparsers.add_parser('build',
+        help="Build wheel and sdist",
+    )
+
     parser_install = subparsers.add_parser('install',
         help="Install the package",
     )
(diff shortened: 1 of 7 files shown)


Carreau:
I think that for "simplicity" you may want to not allow to upload only one format. If you want users can fallback to twine, and upload file individually.

takluyver:
That's true. But I'm thinking that I may at some point deprecate `flit wheel` in favour of `flit build`,  in which case you'd need a way to build just a wheel (for the no-VCS case), and then it's easy to use similar syntax to upload just a wheel.

takluyver:
Suggested alternative names (from discussion on #108) .

Build:
- `flit build`
- `flit package`

Upload:
- `flit upload`
- `flit push`

takluyver:
I'm sticking with `flit build`, but I've changed the other to `flit publish`.

Both commands also take `--format wheel` or `--format sdist`. The parameter can be specified multiple times, so `--format wheel --format sdist` explicitly builds both. Of course, the default if you don't specify either is to build both.

PR #131, opened 2017-08-10
README polishing

alexander255:
Looks like my first PR is not going to be code-related at all. 🙂 

Background: I also found `flit`'s README underwelming and think that the few attached tweaks make the package *look* a lot more interesting, while also improving readablity. The project's SVG image was exported as PNG because [Github won't have it otherwise](https://stackoverflow.com/questions/13808020/include-an-svg-hosted-on-github-in-markdown).

Commits 2 & 3 fix minor content-related “issues” and can be omitted if you prefer / disagree.


Carreau:
+1, though I think more and more, the pip executable is pip for Python 3.

alexander255:
@Carreau: If you have both versions installed (PIP for Python2 and PIP for Python3), then the system still defaults to PIP being Python 2 on Debian and Fedora (Linux) due to back-compat considerations.

README.rst
```diff
@@ -1,7 +1,26 @@
+flit
```
takluyver: The README is included as the front page of the docs, so please leave off this title. I prefer READMEs to open with text anyway.

We'll probably need to remove the logo from the docs so that it's not there twice, as well.

README.rst
```diff
@@ -1,7 +1,26 @@
...
+====
+
+.. figure:: logo.png
+   :alt: The Flit Python packaging tool
```
takluyver: Can we have the logo on the right with text flowing around it? I think adding `:align: right` should do it.

takluyver:
I go back and forth on the `pip` vs `pip3` question. Yes, the default on Linux distros is for `pip` to still mean Python 2, but in conda environments you only have `pip`, there's no `pip3` or `pip2` command. So there's nothing that works for everyone.

I think I'd rather leave it as `pip` and let people work out which pip they need to invoke. There is a clear note that it requires Python 3.

takluyver:
Sorry it took me a while to look at this, by the way - I've been travelling.

README.rst
alexander255: No, because GitHub doesn't support this either. See this similar issue for `asciidoc`: https://github.com/github/markup/issues/984

Basically what happens is:

  1. GitHub renderer calls `rest2html` on the `README.rst` file
  2. The HTML output now contains:
```html
<div class="figure align-right">
<img alt="The Flit Python packaging tool" src="logo.png" />
</div>
 ```
   3. The HTML is nuked through GitHub's HTML Pipeline [sanitization filter](https://github.com/jch/html-pipeline/blob/master/lib/html/pipeline/sanitization_filter.rb), which strips off all `style` definitions and `class` attributes, resulting in this HTML markup actually being renderer:
```html
<div>
<img alt="The Flit Python packaging tool" src="logo.png" />
</div>
 ```

README.rst
alexander255: > The README is included as the front page of the docs, so please leave off this title. I prefer READMEs to open with text anyway.

Althrough I don't necessarily agree on this sentiment, I'm not gonna argue about this. Fixed.

> We'll probably need to remove the logo from the docs so that it's not there twice, as well.

I'm not sure what you're saying, you should definitely leave a vector-graphics version of your logo around in any case.

alexander255:
> I go back and forth on the pip vs pip3 question. Yes, the default on Linux distros is for pip to still mean Python 2, but in conda environments you only have pip, there's no pip3 or pip2 command. So there's nothing that works for everyone.

I thought it might be easier for inexperienced devs (I definietly would have fallen victim to this a couple of years ago), but if you say that it doesn't work universially using the `pip3` command, then it's probably fine to leave it at being just `pip`.
How about adding a note that the command may be `pip3` in some environments instead?

> Sorry it took me a while to look at this, by the way - I've been travelling.

No problem. 🙂

README.rst
takluyver: The docs page which includes this already has the logo in: http://flit.readthedocs.io/en/latest/

If the README includes the logo as well, the docs would have it twice.

But since we can't right-align it in the README, I'd rather drop the logo from the README so that the useful text starts sooner, and leave it just in the docs, where we can right-align it.

takluyver:
> How about adding a note that the command may be pip3 in some environments instead?

Yup, that sounds good. It can be part of the note about requiring Python 3.

README.rst
alexander255: *Since you seem to dislike the idea of adding the logo to the README in general*:
The reason I've put it there in the first place is because the logo is great for capturing the interest of potential users. Yes, this is a programming tool not an end-user app, but in a lot of ways this tool is also a statement that “Python packaging can be simple”, which the logo IMHO conveys pretty nicely. Finally it's just nice to have a little bit of visual change in the otherwise pretty black-and-white, matter-of-factly documentation that we programmers always stare at. It really adds a lot.
(Remember: We're talking about adding 150px of visual information at the top of a `README` here, not about addind a page-size flicking advertisment.)

> If the README includes the logo as well, the docs would have it twice.

I don't see how having the logo in the README means that the logo cannot be also be visible on the documentation?

takluyver: Sorry. I like the added colour of the logo, and if it could be floated to the right so it wasn't pushing the text further down, I'd be happy to have it in the readme. But as it can't, my preference is to leave it out and put up with the readme looking a bit dull.

takluyver:
Thanks for your patience on this - I know I'm asking you to revert quite a bit of what you've done, and that's got to be frustrating.

When you get down to it, the README is actually already quite close to what I want it to be. It looks rubbish to you because we have different priorities, not because I don't care about it! 😉

brettcannon:
Actually the best way to make sure people use the proper interpreter with pip is `python3 -m pip` (or whatever interpreter one wants to use) and completely ignore the whole `pip`/`pip3` discussion.

takluyver:
Thanks Brett, that's a good point. I think I'll change the README to invoke it with `python3 -m`.

PR #136, opened 2017-10-08
First steps towards reproducible sdist

takluyver:
This doesn't make sdists properly reproducible, but if you set `SOURCE_DATE_EPOCH` and build twice on the same machine (so the files have the same mtimes and permissions), you can get the same result.

CC @Carreau - we were talking about this somewhere else. I think this is all that's necessary to set the gzip timestamp:

```python
source_date_epoch = os.environ.get('SOURCE_DATE_EPOCH', '')
mtime = int(source_date_epoch) if source_date_epoch else None
gz = GzipFile(str(target), mode='wb', mtime=mtime)
tf = tarfile.TarFile(str(target), mode='w', fileobj=gz)
```

### DIFF
--- a/flit/sdist.py
+++ b/flit/sdist.py
@@ -1,5 +1,6 @@
 from configparser import ConfigParser
 from collections import defaultdict
+from copy import copy
 from gzip import GzipFile
 import io
 import logging
@@ -124,6 +125,25 @@ def include_path(p):
                 or (os.sep+'__pycache__' in p)
                 or p.endswith('.pyc'))
 
+def clean_tarinfo(ti, mtime=None):
+    """Clean metadata from a TarInfo object to make it more reproducible.
+
+    - Set uid & gid to 0
+    - Set uname and gname to ""
+    - Normalise permissions to 644 or 755
+    - Set mtime if not None
+    """
+    ti = copy(ti)
+    ti.uid = 0
+    ti.gid = 0
+    ti.uname = ''
+    ti.gname = ''
+    ti.mode = common.normalize_file_permissions(ti.mode)
+    if mtime is not None:
+        ti.mtime = mtime
+    return ti
+
+
 class SdistBuilder:
     def __init__(self, ini_path=Path('flit.ini')):
         self.ini_path = ini_path
@@ -222,7 +242,14 @@ class SdistBuilder:
 
             for relpath in files_to_add:
                 path = self.srcdir / relpath
-                tf.add(str(path), arcname=pjoin(tf_dir, relpath))
+                ti = tf.gettarinfo(str(path), arcname=pjoin(tf_dir, relpath))
+                ti = clean_tarinfo(ti, mtime)
+
+                if ti.isreg():
+                    with path.open('rb') as f:
+                        tf.addfile(ti, f)
+                else:
+                    tf.addfile(ti)  # Symlinks & ?
 
             if 'setup.py' in files_to_add:
                 log.warning("Using setup.py from repository, not generating setup.py")
--- a/flit/sdist.py
+++ b/flit/sdist.py
@@ -1,5 +1,6 @@
 from configparser import ConfigParser
 from collections import defaultdict
+from gzip import GzipFile
 import io
 import logging
 import os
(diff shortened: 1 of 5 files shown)


Carreau:
Thanks for pinging me ; looking into it !

flit/sdist.py
```diff
@@ -209,37 +210,44 @@ def build(self, target_dir:Path =None):
...
+            for relpath in files_to_add:
+                path = self.srcdir / relpath
+                tf.add(str(path), arcname=pjoin(tf_dir, relpath))
+
+            if 'setup.py' in files_to_add:
+                log.warning("Using setup.py from repository, not generating setup.py")
+            else:
+                setup_py = self.make_setup_py()
+                log.info("Writing generated setup.py")
+                ti = tarfile.TarInfo(pjoin(tf_dir, 'setup.py'))
```
Carreau: I believe you might have to set the file creation of setup.py here to SDE.

flit/sdist.py
```diff
@@ -209,37 +210,44 @@ def build(self, target_dir:Path =None):
...
+
+            pkg_info = PKG_INFO.format(
+                name=self.metadata.name,
+                version=self.metadata.version,
+                summary=self.metadata.summary,
+                home_page=self.metadata.home_page,
+                author=self.metadata.author,
+                author_email=self.metadata.author_email,
+            ).encode('utf-8')
+            ti = tarfile.TarInfo(pjoin(tf_dir, 'PKG-INFO'))
```
Carreau: Maybe same here ?

Carreau:
There might be two places where you have to set the mtime explicitly, but otherwise +1

takluyver:
Just went to investigate the `tarfile` module. It looks like it uses 0 as the default mtime when there isn't a file to get the mtime from (whereas wrapping the tarball into a gz uses the current time).

I'll still need to figure out overriding the mtime of files going into the tarball to make a properly reproducible sdist, but I have tried this locally and I could build an identical sdist twice on the same computer.

Carreau:
> Just went to investigate the tarfile module. It looks like it uses 0 as the default mtime when there isn't a file to get the mtime from

Got you, I missed it was in memory, the `pjoin()` made me think it was written on disk :-) 

I'm confused  from the code, aren't the setup.py

> I'll still need to figure out overriding the mtime of files going into the tarball to make a properly reproducible sdist, but I have tried this locally and I could build an identical sdist twice on the same computer.

Yeah, that's annoying, you need to get the tarInfo and overwrite the mtime/atime attribute. 

If you can reproduce on same computer then +1 already, we can figure out the rest later.

flit/sdist.py
```diff
@@ -209,37 +210,44 @@ def build(self, target_dir:Path =None):
...
-        tf.close()
+        source_date_epoch = os.environ.get('SOURCE_DATE_EPOCH', '')
+        mtime = int(source_date_epoch) if source_date_epoch else None
+        gz = GzipFile(str(target), mode='wb', mtime=mtime)
+        tf = tarfile.TarFile(str(target), mode='w', fileobj=gz)
+
+        try:
+            tf_dir = '{}-{}'.format(self.metadata.name, self.metadata.version)
+
+            files_to_add = self.find_tracked_files()
```
techtonik: Is that guaranteed to return sorted files?

takluyver: Yup:

https://github.com/takluyver/flit/blob/8f5d6a71a5bc02ac0a0ba861059a5dd71d407829/flit/sdist.py#L165

Good thing to check, though :-)

takluyver:
OK, I've just built identical tarballs on two different machines, one Fedora and one Ubuntu. Who wants to test this some more?

1. Check out this branch, so the last commit is fa12529
2. Set SDE: `export SOURCE_DATE_EPOCH=1507470746`
3. `flit build`
4. `sha256sum dist/flit-0.11.4*`

Result should be like this:

```
460794e84499cd679da945fc6987d9447b8d9829625fae3fd837eadfe1770ddc  dist/flit-0.11.4-py3-none-any.whl
12060d1cee6bbe704d57495675d8f45ecd5e0fe6406c45cca6474106eb623f92  dist/flit-0.11.4.tar.gz
```

If either the wheel or the sdist doesn't match, send me a copy so I can investigate the difference!

Carreau:
I confirm I got the same, and changing SDE affect the hash of the sdist, not the wheel. 
```
460794e84499cd679da945fc6987d9447b8d9829625fae3fd837eadfe1770ddc  dist/flit-0.11.4-py3-none-any.whl
12060d1cee6bbe704d57495675d8f45ecd5e0fe6406c45cca6474106eb623f92  dist/flit-0.11.4.tar.gz
```

Carreau:
Should I merge, or do you want anything more ?

takluyver:
Nice! Weird that sde didn't change the hash of the wheel, though. Might
investigate what's going on there.

I'll have a look tomorrow for tests I can easily add, but then I think this
will be ready to merge.

flit/sdist.py
techtonik: Nice. =) I am just thinking if explicitly sorting here makes the code more readable.

takluyver:
OK, I just rebuilt with a different SOURCE_DATE_EPOCH, and both the sdist and the wheel changed. So I think that might have been an illusion that the wheel didn't. :-P

PR #139, opened 2017-10-10
Ignore empty lines in all metadata list fields

zenbot:
This addresses https://github.com/takluyver/flit/issues/129.


takluyver:
Thanks!

PR #143, opened 2017-10-26
Extract a package's summary & version via ast

zenbot:
Per discussion in https://github.com/takluyver/flit/issues/141, this POC PR extracts a package under build's version & summary via `ast`, falling back to an import if that fails (because, for example, the `__version__` is anything but a simple string assignment).

This change allows packages to be built in an environment where the package being built can't necessarily be imported.

### DIFF
--- a/flit/common.py
+++ b/flit/common.py
@@ -1,3 +1,4 @@
+import ast
 from contextlib import contextmanager
 import hashlib
 from importlib.machinery import SourceFileLoader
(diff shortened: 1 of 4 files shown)


takluyver:
Thanks! I'll turn this over in my head for a few more days while I'm at a conference, but I it looks good - nice clean implementation, and tests to check it both ways.

takluyver:
I'm going for this, thanks.

PR #145, opened 2017-10-29
add option to use pth file on install command

aodag:
`flit install -s` uses symlink, so this option can't work on windows.
this PR adds option using pth file to add project directory to `sys.path`.

### DIFF
--- a/flit/install.py
+++ b/flit/install.py
@@ -85,10 +85,11 @@ class RootInstallError(Exception):
 
 class Installer(object):
     def __init__(self, ini_path, user=None, python=sys.executable,
-                 symlink=False, deps='all'):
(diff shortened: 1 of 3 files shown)


flit/install.py
```diff
@@ -266,6 +271,13 @@ def install_directly(self):
 
         self.write_dist_info(dirs['purelib'])
 
+    def add_pth_entry(self, src, dst):
+        log.info("pth %s %s", src, dst)
```
takluyver: This log call is superfluous - the same info is logged just before.

takluyver: Since this is a simple thing, I'm going to merge it and remove the log call myself. Thanks for your contribution!

PR #146, opened 2017-11-08
Bring back zipfile36 as a requirement for older Pythons

zenbot:
It looks like this may have been removed unintentionally when flit's own build config moved to `pyproject.toml` — it's definitely still required at runtime on Pythons < 3.6.


takluyver:
Thanks, you're absolutely right :-)

takluyver:
I've done a 0.12.1 release with this fix.

PR #147, opened 2017-11-14
Tool to convert flit.ini to pyproject.toml

takluyver:
In a folder containing `flit.ini`, run:

```shell
python3 -m flit.tomlify
```

I haven't currently given this a subcommand, because it's a rougher temporary tool.

### DIFF
--- a/doc/pyproject_toml.rst
+++ b/doc/pyproject_toml.rst
@@ -9,6 +9,9 @@ This file lives next to the module or package.
    similar information. Flit can still read these files for now, but you should
    switch to ``pyproject.toml`` soon.
 
+   Run ``python3 -m flit.tomlify`` to convert a ``flit.ini`` file to
+   ``pyproject.toml``.
+
 Build system section
 --------------------
 
--- a/flit/init.py
+++ b/flit/init.py
@@ -189,7 +189,7 @@ class TerminalIniter(IniterBase):
         print()
         print("Written pyproject.toml; edit that file to add optional extra info.")
 
-TEMPLATE = """
+TEMPLATE = """\
 [build-system]
 requires = ["flit"]
 build-backend = "flit.buildapi"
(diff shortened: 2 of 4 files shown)


Carreau:
I'll test that on all my flit projects. Damn you are active. I was willing
to do something like that at some point.

takluyver:
Ah, if I'd known it was on your radar I might have left it. I was about to convert one of my projects by hand, and I thought...

develop a system for passing arbitrary condiments

Carreau:
> and I thought...

Exactly ! Well now we can search GitHub for `flit.ini` and automatically send PRs !

Mariatta:
Cool! I will try this on [pytaco](https://github.com/mariatta/pytaco) later 😀

takluyver:
Let's hold off on sending lots of PRs for the time being. Automated PRs are annoying, I haven't tested the conversion extensively, and I'd like to give people a chance to upgrade of their own accord before pushing them.

Carreau:
AFAICT no issue for converting my [there](https://pypi.io/project/there/) package. Assuming every project are under version control you could also remove flit.ini.

takluyver:
Great, thanks for testing it.

I figure we should let people remove `flit.ini` themselves, just in case there's a bug in the conversion and it's somehow not in version control. Or you might want to check the new file manually against the old one. Smart tools can be too smart. ;-)

Carreau:
+1 project converted (so far no issues). Should `[build-system]requires = ["flit"]`  enfoce a minimum version that knows about pyproject.toml ?

takluyver:
Could do, but the way that pip handles build system requirements, I think it will always install the newest one compatible with the requirements.

Mariatta:
Perhaps I should have reported this update earlier: I've successfully converted 3 projects from flit.ini to pyproject.toml using `flit.tomlify` :)

takluyver:
Excellent :-)

PR #154, opened 2017-12-08
More validation of packaging data

takluyver:
This should make it easier to catch mistakes in packaging sooner.

The design is slightly awkward: a series of validation functions build up a list of problems. These are logged at 'error' level, and then a ConfigError is raised. I'm doing it this way so that it's easy to see multiple problems at once, not just the first one it finds.

### DIFF
--- a/doc/cmdline.rst
+++ b/doc/cmdline.rst
@@ -139,6 +139,19 @@ Environment variables
    See :ref:`uploading packages with environment variables <upload_envvars>`
    for more information.
 
+.. envvar:: FLIT_ALLOW_INVALID
+
+   .. versionadded:: 0.13
+
+   Setting this to any non-empty value tells Flit to continue if it detects
+   invalid metadata, instead of failing with an error. Problems will still be
+   reported in the logs, but won't cause Flit to stop.
+
+   If the metadata is invalid, uploading the package to PyPI may fail. This
+   environment variable provides an escape hatch in case Flit incorrectly
+   rejects your valid metadata. If you need to use it and you believe your
+   metadata is valid, please `open an issue <https://github.com/takluyver/flit/issues>`__.
+
 .. envvar:: SOURCE_DATE_EPOCH
 
    To make reproducible builds, set this to a timestamp as a number of seconds
--- a/flit/validate.py
+++ b/flit/validate.py
@@ -7,6 +7,8 @@ import re
 import requests
 import sys
 
+from .common import InvalidVersion
+
 log = logging.getLogger(__name__)
 
 def get_cache_dir() -> Path:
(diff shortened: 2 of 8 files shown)


takluyver:
This is taking shape. I have added an environment variable `FLIT_ALLOW_INVALID`. If Flit erroneously rejects your metadata, you can use this as an escape hatch:

```shell
FLIT_ALLOW_INVALID=1 flit build
```

takluyver:
@Carreau I've extended the PEP 440 checks on the version number. Now it tries to normalise the version to the canonical PEP 440 form, with an error if it can't do that, and a warning if it can do that and it has to change the version.

doc/cmdline.rst
```diff
@@ -139,6 +139,19 @@ Environment variables
    See :ref:`uploading packages with environment variables <upload_envvars>`
    for more information.
 
+.. envvar:: FLIT_ALLOW_INVALID
```
Carreau: `FLIT_ALLOW_INVALID_METADATA` ? You may want invalid other things later on ?

flit/validate.py
```diff
@@ -0,0 +1,335 @@
...
+def get_cache_dir() -> Path:
+    """Locate a platform-appropriate cache directory for flit to use
+
+    Does not ensure that the cache directory exists.
+    """
+    if os.name == 'posix' and sys.platform != 'darwin':
+        # Linux, Unix, AIX, etc.
+        # use ~/.cache if empty OR not set
+        xdg = os.environ.get("XDG_CACHE_HOME", None) or (
+        os.path.expanduser('~/.cache'))
```
Carreau: Hard to read indentation (IMHO)

flit/validate.py
```diff
@@ -0,0 +1,335 @@
...
+    return ["Unrecognised classifier: {!r}".format(c)
+            for c in sorted(invalid)]
+
+
+def _download_classifiers():
+    """Get the list of valid trove classifiers from PyPI"""
+    log.info('Fetching list of valid trove classifiers')
+    resp = requests.get(
+        'https://pypi.python.org/pypi?%3Aaction=list_classifiers')
+    resp.raise_for_status()
```
Carreau: possibly find the corresponding URL for warehouse, and use as a fallback ?

doc/cmdline.rst
takluyver: I think I'll live with that risk. I'm not sure that the scope of 'metadata' is very clear, e.g. entry points are not part of the `[tool.flit.metadata]` table.

This is meant to be an escape hatch for when a check is wrong, so I don't think it needs to be fine-grained.

flit/validate.py
takluyver: Is the new commit better? I'm not quite sure what was making it tricky.

flit/validate.py
takluyver: I've found https://pypi.org/pypi?%3Aaction=list_classifiers , but I'm asking whether it's something we can rely on before I switch to it: https://github.com/pypa/warehouse/issues/1300

takluyver: Donald says it's safe to rely on, so I've switched to that.

Carreau: Good !

doc/cmdline.rst
Carreau: I'm ok with it, it was mostly a question. You're the boss, boss.

flit/validate.py
Carreau: It's just no indentation after the `or (\n`, looks to my eyes (brain?), like a missing assignment. At least I don't easily see the it's part of the above assignment. It may be britany's brain flakes are of low quality.

doc/cmdline.rst
takluyver: who's the boss

It was a point worth thinking about :-)

flit/validate.py
takluyver: Ah, I see it now. It was so misleading I didn't even notice what it was meant to be ;-)

takluyver:
Thanks for the review :-) . Do you want to look any more at it, or shall one of us press the big green button?

Carreau:
I'll give another read in a few hours if you wish, but otherwise, let's
test it !

takluyver:
I've just tried it with a bunch of projects I've got on this machine, and it wasn't causing any problems, so I'm going to merge it now. I'll give it a day or two in master before making a new release.

PR #155, opened 2017-12-12
Remove mention of flit.ini as default file from readme.

Carreau:
Also rename `docs` to `documentation`

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -49,8 +49,8 @@ or as a directory — and you want to distribute it.
 
    You can edit this file to add other metadata, for example to set up
    command line scripts. See the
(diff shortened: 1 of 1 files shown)


takluyver:
Good catch!

Carreau:
Thinking of starting to write a PyCon proposal. So far I have the title: "Packaging like it's '18" play on 2018/pep-518. Deadline is IIRC Jan 4 for PyCon.

takluyver:
Sounds like a good idea, though I'm not sold on the title. I didn't get the reference to either the year or the PEP. And the number 18 has a range of connotations because it's traditionally the age when people get a new set of rights and responsibilities.

I haven't come up with anything I really like, but maybe just expand the abbreviation: "Packaging like it's 2018".

PR #159, opened 2018-01-16
WIP: added supporting namespace package

aodag:
- [won't fix] support pkg_resources style namespace package (adding namespace_packages in setup.py)
- [x] fix installing via symlink
- [x] refactor variable names


takluyver:
Thanks for this. I'm really impressed that adding support for namespace packages takes so small a change. Have you checked manually that the wheels and sdists it builds are right?

flit/common.py
```diff
@@ -14,12 +15,20 @@ class Module(object):
...
 
         # It must exist either as a .py file or a directory, but not both
+        name = (name + ".").split(".", 1)[0]
         pkg_dir = Path(directory, name)
         py_file = Path(directory, name+'.py')
         if pkg_dir.is_dir() and py_file.is_file():
             raise ValueError("Both {} and {} exist".format(pkg_dir, py_file))
+        elif pkg_dir.is_dir() and "." in self.name:
+            self.path = pkg_dir
+            self.ns_path = Path(directory, self.name.replace('.', os.sep))
```
takluyver: I might request that some of these names change a bit - to me, the 'namespace path' would refer to the outer directory, not the inner one. But I haven't worked out what names I want yet.

The code should also check that the inner directory exists - currently I think your changes make it possible to instantiate this object when the directory is missing, which wasn't possible before.

aodag:
I checked creating sdist and wheel for native namespace package, and that works fine.

Supprting pkg_resources namespace requires adding `namespace_packages` entry in `setup.py`.
I'm also going to add this feature.

* https://www.python.org/dev/peps/pep-0420/
* https://packaging.python.org/guides/packaging-namespace-packages/#native-namespace-packages

flying-sheep:
hi @takluyver any chance this gets merged soon?

takluyver:
What does this need doing?

- Better variable naming to make the code clearer (see my comment above)
- The code that finds the module needs to be more robust against invalid combinations 
  - e.g. if you specify `module=foo.bar`, it could currently package up `foo.py` and ignore the 'bar' part entirely.
- @aodag mentioned that it doesn't yet work with installing by symlink.
- We also need to check it with the `--pth-file` option.
- I don't want to support the pre-standardisation (pkgutil or pkg_resources) namespace packages, so the test & sample for pkgutil should go. People who need to do namespace packages for Python 2 can stick with setuptools.
- Probably a few more tests, including building wheels and sdists

flying-sheep:
@aodag did you implement all the things @takluyver mentioned? I’d like to package a sphinx extension, and since flit exists, using setup.py kinda hurts.

takluyver:
I was hoping to look at this properly before releasing 1.1, but I think I'm unfairly holding up the release waiting to get time for that, so I'm going to try to get the release out and come back to it.

At a brief glance, this is getting closer, and I'm still pleased to see how few changes it seems to need, but I think it could be a bit clearer.

flying-sheep:
Hi! Should I have a go at the tests or are you making progress?

takluyver:
I'm not quite sure what to do with this; I've got it 'working' in the sense that the tests pass and I'm reasonably happy with the code that's there. But I'm not confident that it will reliably do the right thing in all the different cases (install with pip, install symlink, wheel, sdist with setup.py, sdist with PEP 517...).

I ~never use namespace packages myself; my experiences with them have always been that they're an unnecessary complication, creating new ways for things to go wrong for little benefit. So I don't have much motivation to do all the testing to satisfy myself that this will work properly. And I don't have much enthusiasm for maintaining it if it doesn't work.

flying-sheep:
They are needed if you want to install things inside of a module hierarchy (`extensible_package.ext.myplugin`).

Namespace packages are a part of python and whenever a tool doesn’t support them and breaks in interesting and hard to debug ways when encountering them, I blame the tool, not namespace packages. They are innocent: They’re just directories! I think it’s rather straightforward to support them if they aren’t treated as an afterthought.

takluyver:
For most of their history, though, they've been directories with a special marker file that does something hardly anyone really understands. And then they break in hard to understand ways. I spent ages trying to debug a problem which affected IPython when you installed one `backports.<something>` package and it broke another one. It could all have been avoided if they had just been `backports_<something>` instead.

Now they're 'just a directory', but that can cause its own problems, if you try/except an import and it succeeds because some random directory is found as a namespace package. And I've no idea if new-style namespace packages play well with old-style namespace packages.

So, in summary, I think things would be a bit simpler if namespace packages had never been invented and we just had to use `_` as a delimiter in package names.

flying-sheep:
> For most of their history, though, they've been directories with a special marker file that does something hardly anyone really understands. And then they break in hard to understand ways.

Yes. Therefore I also think that old-style namespace packages should die.

> Now they're 'just a directory', but that can cause its own problems, if you try/except an import and it succeeds because some random directory is found as a namespace package.

I had that problem a few times, but never with namespace packages. It was always someone fucking up by e.g. installing `test` to site-packages by using `find_packages()`. Prepending to `PYTHONPATH` should also be done with care.

> So, in summary, I think things would be a bit simpler if namespace packages had never been invented and we just had to use _ as a delimiter in package names.

I think you’d think differently if new-style namespace packages had been available forever. If Guido in 1991 had said something like

> Python packages are directories that may have python files in them. You can optionally use `__init__.py` to write initialization code.

This is the way it is now, and I think it’s a rather simple concept hampered by lack of tool support for that kind of packages.

rupertnash:
So just resurrecting this in the hope of this feature getting merged in the near future

I have to agree with @aodag: 
> Namespace packages are a part of python and whenever a tool doesn’t support them and breaks in interesting and hard to debug ways when encountering them, I blame the tool, not namespace packages. 

I have `foo.core`, `foo.client` and `foo.server` packages stored in three separate source trees, and their build/install needs to be cleaned up. Ideally I'd use flit but without the namespace support I'm stuck for now.

Afraid I'm not knowledgable enough about all the corner cases in python packaging to do more than update the fix-namespace branch with current master, which I've done on my fork: https://github.com/rupertnash/flit/tree/fix-namespace

Tests pass except for py35 which dies at package install time:
```
action: py35, msg: installpkg
cwd: /Users/rnash2/work/vestec/flit-update/flit
cmd: /Users/rnash2/work/vestec/flit-update/flit/.tox/py35/bin/python -m pip install --exists-action w .tox/.tmp/package/1/flit-1.3.tar.gz
Processing ./.tox/.tmp/package/1/flit-1.3.tar.gz
  Missing build time requirements in pyproject.toml for file:///Users/rnash2/work/vestec/flit-update/flit/.tox/.tmp/package/1/flit-1.3.tar.gz: 'setuptools' and 'wheel'.
  This version of pip does not implement PEP 517 so it cannot build a wheel without 'setuptools' and 'wheel'.
  Could not find a version that satisfies the requirement pytoml (from versions: )
No matching distribution found for pytoml
You are using pip version 10.0.1, however version 19.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
```

maggyero:
@takluyver Any chance this PR get merged? We need this to adopt Flit at my company, as we have several regular subpackages of a company _namespace_ package spread across several distributions:

```
.git/
company/
|-- subpackage_a/
    |-- __init__.py
tests/
pyproject.toml
```

```
.git/
company/
|-- subpackage_b/
    |-- __init__.py
tests/
pyproject.toml
```

Client code:

```
from company.subpackage_a import object_x
from company.subpackage_b import object_y
```

We should not use ugly prefixes like in C when we have namespace support like in C++ or Python in my opinion.

maggyero:
@takluyver I saw that you have recently merged the `src` layout PR (https://github.com/takluyver/flit/pull/260) and plan to make a new release soon (Flit 2.0). Do you plan to merge this PR as well for the new release? It would be awesome.

stefco:
@takluyver Are there any plans to merge this?

I'll also add that, like properties/descriptors, they are useful for preserving interfaces/allowing dependent code to be unchanged. Flit is great, and I'd really like to be able to use it for the subcomponents of my project that I'm splitting out. 

I'm happy to help if there's something that needs to be done to make this work.

takluyver:
Sorry that I've been letting this linger for ages. I'll accept that namespace packages are a thing lots of people want, even if I'm not a fan myself.

I've just commented in #453 that I prefer the approach in this PR - where you specify that you're packaging the concrete subpackage rather than the shared namespace package. However, this PR no longer merges, due to major refactorings that have happened in the meantime. That's entirely my fault for ignoring it for so long.

Could I ask someone to port this model to the new code structure? Whether that's a rebase of this branch, building on @martenlienen's work, or a new branch entirely. I think most of the changes in this branch will simply go to the corresponding files in `flit_core` (which is part of this repository), but there may well be bits that require some more thought. I will do my best to get this in before too much time passes.

(Just in case there's any doubt, we'll only do the standardised [PEP 420](https://www.python.org/dev/peps/pep-0420/) namespace packages, not the older style. That was my preference already back in 2018, and hopefully by now it's entirely uncontroversial.)

tests/samples/ns1-pkg2/ns1-pkg2.toml
```diff
@@ -0,0 +1,9 @@
+[build-system]
+requires = ["flit"]
+
+[tool.flit.metadata]
+module = "ns1.pkg2"
+author = "Sir Robin"
+author-email = "robin@camelot.uk"
+home-page = "http://github.com/sirrobin/module1"
```
merwok, line 8: sirrobin is a real github account!

Probably better to use canonical `example.com` domain for these

takluyver: Yes, that's a good idea. I think the 'Sir Robin' example is copied and pasted around the tests.

takluyver:
#468 is a rebase of this.

PR #160, opened 2018-01-17
Remove subcommands: wheel, sdist, register

takluyver:
These were deprecated in 0.13.

* *register* doesn't work with Warehouse, and since legacy PyPI no longer accepts uploads, it doesn't seem important to keep it. Users should now just upload packages directly.
* `flit build` now produces wheels and sdists, so the separate commands for them are no longer needed.

### DIFF
--- a/flit/__init__.py
+++ b/flit/__init__.py
@@ -88,10 +88,6 @@ def main(argv=None):
         help="Prepare flit.ini for a new package"
     )
 
-    subparsers.add_parser('register',
-        help="register a package on PyPI without uploading any files"
-    )
-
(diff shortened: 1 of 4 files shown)


Carreau:
Less code means less bugs !

Carreau:
Do you want to add that in the changelog, or merge as is, and do the changelog at release time ?

takluyver:
I'll do the changelog when I'm getting ready to release. I'm planning some refactoring of the subcommand machinery, but I thought I'd leave it until after dropping the older subcommands.

PR #161, opened 2018-01-18
Restructure how subcommands work

takluyver:
I'm not convinced that this is actually an improvement - it looked better in my imagination - but having done it I thought I'd open a PR to consider it.

In master, all the command line parsing code is in one place, and it dispatches to individual bits of functionality in other modules. This PR puts the command line parsing for each subcommand into the module with the code it calls.

Pros:

- Arguments are associated with the code that uses them.
- Nicer API for defining subcommands.
- Slightly nicer help message (I spent a while tweaking it to be just how I want).

Cons:

- CLI definition is no longer together in one place.
- More code to maintain.
- Some bits use private attributes of `argparse` objects.

### DIFF
--- a/flit/__init__.py
+++ b/flit/__init__.py
@@ -1,91 +1,45 @@
 """A simple packaging tool for simple packages."""
 import argparse
 import logging
-import pathlib
+from pathlib import Path
 import sys
 
 from . import common
 from .log import enable_colourful_output
+from .subcmds import Subcommand, SubcommandArgumentParser
 
 __version__ = '0.13'
 
 log = logging.getLogger(__name__)
 
-def add_shared_install_options(parser):
-    parser.add_argument('--user', action='store_true', default=None,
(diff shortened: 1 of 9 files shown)


djmoch:
I don’t know if you’re looking for comments or not, but I’ll offer some feedback anyway in the hope that it’s useful. 

In short, if your goal is to separate your main entry point logic from the argument parsing logic, I think a separate `cli.py` file would make more sense. Mixing the argument parsing into the “business logic” just creates new confusion in place of the old.

takluyver:
Thanks. My thinking was to create a nice framework for defining subcommands, because I don't much like the way argparse does it. But when I actually implemented it, I didn't like the way it spread argument parsing stuff into different modules.

So this PR isn't going anywhere in its current state, but it's lingering as a kind of placeholder in case I get another idea about subcommand handling.

PR #169, opened 2018-04-04
Official support for markdown and plain text descriptions

### LINKED ISSUE #166: Support publishing markdown descriptions to Warehouse
PEP 566 is now implemented in Warehouse:
https://dustingram.com/articles/2018/03/16/markdown-descriptions-on-pypi

takluyver:
Fixes #166.

Until recently, PyPI only supported rst descriptions, falling back to plain text if the rst was invalid. Now a new metadata field `Description-Content-Type` lets us specify whether the description is rst, markdown or plain text. This new code uses that, picking the content type based on the file extension (`.rst`, `.md` or `.txt`). If the extension is something else, Flit will warn the user and leave the field unset, so it's up to PyPI what to do with it.

Markdown support has been a secret feature of Flit for some time, by using pandoc to convert markdown to rst (thanks @Carreau!). This makes it an official feature, and gets rid of the code to call pandoc, since PyPI can now handle markdown natively.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ __pycache__/
 /tests/samples/dist/
 /htmlcov/
 /.coverage
+/.pytest_cache
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ python:
   - "3.5"
   - "3.4"
 install:
+  - pip install --upgrade pytest
   - pip install -r requirements-test.txt
   - pip install codecov
 script: py.test --cov=flit
--- a/doc/history.rst
+++ b/doc/history.rst
@@ -4,6 +4,10 @@ Release history
 Version 0.14
 ------------
 
+- The description file may now be written in reStructuredText, Markdown or
+  plain text. The file extension should indicate which of these formats it is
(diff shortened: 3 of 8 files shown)


Carreau:
There is now a [variant type](https://packaging.python.org/specifications/core-metadata/#description-content-type-optional) being GFM (default. GitHub FLavored Markdown), or CommonMark.

I don't think it will import (much), but we should have an option to set the variant. 

I' havent dive in the code, runnign to lunch.

takluyver:
This is partly laziness, but I'm inclined to ship it without that option (i.e. all markdown will be GFM for now), and add it later if people want it.

Carreau:
The `charset` seem to also be an option I'm in favor of enforcing utf-8 and/or adding a markdown-variant flag in a subsequent PR.

takluyver:
The spec says about charset:

> The only legal value is UTF-8. If omitted, it is assumed to be UTF-8.

We already read the file as UTF-8, so it will fail if you point it to a file that isn't valid UTF-8. So I don't think there's anything to do on that front.

PR #183, opened 2018-06-13
Fix examples in doc/flit_ini.rst

Zaab1t:
Hello Thomas Kluyver,

I think `flit` is a great tool, thank you!
This old doc file hasn't kept up with the others, so it bit me when I missed the note at the top. The lists in it are not valid at least according to the parsing lib `pytoml`. Also added spaces around `=`-operators because you do that in your `pyproject.toml` and it looks more like Python :)

Have a nice day,
Carl

### DIFF
--- a/doc/flit_ini.rst
+++ b/doc/flit_ini.rst
@@ -30,10 +30,10 @@ e.g. for flit itself
 .. code-block:: ini
 
     [metadata]
-    module=flit
-    author=Thomas Kluyver
-    author-email=thomas@kluyver.me.uk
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks Carl - there's actually a new file for the new format:

https://github.com/takluyver/flit/blob/master/doc/pyproject_toml.rst

I'm keeping the old documentation page around for now because the `flit.ini` format still works, but there's a note on it about switching to pyproject.toml. So we don't need to update the syntax on that page. :-)

PR #188, opened 2018-07-14
Test against 3.7 dev on travis ci

Mariatta:


Mariatta:
☹️ I believe [bpo-33109](https://bugs.python.org/issue33109) causes the build to fail in Travis' 3.7-dev, but this is fixed in the final release of 3.7.0. (https://github.com/python/cpython/pull/6919).

To test my suspicion, I added `required=False` when calling add_subparsers, and the build now passes in 3.7 (but broke the other versions)

takluyver:
So it sounds like it was fixed in 3.7 before release, and we just need to test on a newer version of 3.7.

takluyver:
I've had a go at this in #191 and it seems to be working - is that OK by you?

Mariatta:
Yup that looks great! Thanks!  Closing in favour of #191

PR #192, opened 2018-07-17
Catch the NoDocstringError when building a package without docstring.

Mariatta:
Closes https://github.com/takluyver/flit/issues/174


takluyver:
I don't understand where the permission error on Windows is coming from. I suspect it's something to do with mocking the same command twice, but we already mock the same command twice and it doesn't seem to have that problem. And I don't understand why it's affecting one Python version and not another.

:confused: 

I tried rerunning the test jobs, but it failed in (as far as I can see) the same place as before. I guess you were also reopening it to rerun the test?

takluyver:
The second error makes some kind of sense after the first - it didn't finish entering the context manager, so it's not being torn down properly. But I'm mystified about the first error at the moment.

Mariatta:
Yeah I re-opened to see if the error might go away the second time 😅 but this seems to be consistent.

I'm confused too why it failed on a specific Python environment. I don't have a windows machine, so I don't know how to debug or fix it..

takluyver:
I'll sleep on it and see if anything occurs to me. If necessary I can reboot my laptop into Windows to investigate.

takluyver:
I rebooted into Windows and ran it with both Python 3.5 and 3.6 (from Miniconda). Tests passed in both. :confused: 

I'll try pressing the button to re-run the tests on Appveyor.

takluyver:
Huh, this time it worked. Weird. Let's try it again to check that working wasn't a fluke...

takluyver:
Nope, it seems like it's happy now. *shrug*

takluyver:
Thanks for contributing :-)

PR #194, opened 2018-07-26
added short option to correspond with `--version`

cajopa:


takluyver:
Seems reasonable, thanks!

PR #195, opened 2018-07-27
Fix the list of projects that 'develop' installs

brettcannon:
It's "doc", not "docs".


takluyver:
Looks like we've caught sporadic test failures on Windows :disappointed:

takluyver:
Thanks

PR #199, opened 2018-08-22
Allow reading project version from source control

webknjaz:
I've promptly integrated setuptools_scm with flit to achieve the ability to keep a version in Git tag.

This is going to be used in python/core-workflow#286

### DIFF
--- a/flit/__init__.py
+++ b/flit/__init__.py
@@ -4,7 +4,7 @@ import logging
 import pathlib
 import sys
 
-from . import common
+from . import common, errors
 from .log import enable_colourful_output
 
 __version__ = '1.0'
@@ -105,7 +105,7 @@ def main(argv=None):
         from .build import main
         try:
             main(args.ini_file, formats=set(args.format or []))
-        except(common.NoDocstringError) as e:
+        except(errors.NoDocstringError) as e:
             sys.exit(e.args[0])
     elif args.subcmd == 'publish':
         from .upload import main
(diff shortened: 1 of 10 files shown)


webknjaz:
Odd, build in my Travis CI account succeeded: https://travis-ci.com/webknjaz/flit/jobs/141478861.
Could you please try rerunning job for 3.4?

webknjaz:
@takluyver ^^^

webknjaz:
Nvmnd, triggered it myself. Ready for review.

takluyver:
I'm glad that you're interested in flit, and I appreciate that you've spent time and effort on this change. However, I'm not going to merge it.

Flit is not designed to support all the different ways people might want to do packaging; it's meant to make it easy and clear to do packaging in one way. That way includes setting the version number in `__version__` inside the module. We often work on the assumption that more options is automatically better, but that's not what I aim for with Flit: it keeps things simple by having one way to do things.

One thing that came out of my work on Flit was a build system interface, defined in PEPs 518 and 517. This is meant to make it more practical for different package building tools to work with tools like pip. So if you want to create a tool that gets the version number from a VCS, it should be more feasible now than it was in the past. (If you do this, please name it so that it's clear it's not part of flit.)

webknjaz:
What about having some hook for plugin in a custom method for version identification? How do change just one part without having to support full fork of flit?
In setuptools there's entry point hooks, for example, which allow other dists to register custom implementations of certain things.

takluyver:
Sorry, at the moment I have no plans for a plugin system, for basically the same reason.

PR #201, opened 2018-08-28
Fix warnings when building the documentation.

Mariatta:
Fix the warning related to `--repository` option.
Add sphinx to requirements-test.
Run `make html` as part of travis CI.

Closes https://github.com/takluyver/flit/issues/189

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,8 @@ install:
   - pip install --upgrade pytest
   - pip install -r requirements-test.txt
   - pip install codecov
-script: py.test --cov=flit
+script:
+  - py.test --cov=flit
+  - sphinx-build -b html -d _build/doctrees doc _build/html
 after_success: codecov
 sudo: false
--- a/doc/upload.rst
+++ b/doc/upload.rst
@@ -60,7 +60,7 @@ Using environment variables
 You can specify a server to upload to with :envvar:`FLIT_INDEX_URL`, and
 pass credentials with :envvar:`FLIT_USERNAME` and :envvar:`FLIT_PASSWORD`.
 Environment variables take precedence over the config file, except if you use
-the :option:`--repository` option to explicitly pick a server from the config file.
+the :option:`flit --repository` option to explicitly pick a server from the config file.
 
 This can make it easier to automate uploads, for example to release packages
 from a continuous integration job.
--- a/requirements-test.txt
+++ b/requirements-test.txt
@@ -8,3 +8,4 @@ pytoml
 pytest>=2.7.3
 pytest-warnings
 pytest-cov
+sphinx


Mariatta:
Hmm, why wouldn't `python3 -m flit install --deps develop` install sphinx, a doc extras?

Carreau:
> Hmm, why wouldn't python3 -m flit install --deps develop install sphinx, a doc extras?

From the logs it logs like it pulls `pyproject.toml` from the pip-installed flit.
Try to bump version in `__init__` to `1.1.dev` ?

Mariatta:
Ah, I see. Do you think 1.1.post0 would make more sense?
Or 1.2.dev0?
I think it should also be its own change, separate from this PR.

Carreau:
Hum, 1.2.dev ? I guess you can try here see if it fixes stuff and then we
can advise. Still strange it uses the pip-installed one... We should try to
figure out why.

Also it's just my guess, I may be wrong !

Mariatta:
Thanks, tested on my machine, and bumping the release number fixed it.

Mariatta:
This is now passing on Travis Python v3.6 and 3.7. I don't know what the problem is with Python 3.5, 3.4. Will look into it later.

Carreau:
It's and installation issue. `zipfile36` does not get installed on 3.4 and 3.5. 
They are not installed anymore. I can try to fix that for you.

Carreau:
All green on travis ! 🥇to @Mariatta !

Appveyor seem to have hit a random error.

Carreau:
> Appveyor seem to have hit a random error.

(and I'm not allowed to restart...)

Mariatta:
Thanks @Carreau. Yeah we've seen that strange error on appveyor before ☹️ 

I was wondering we can move all dependencies from requirements-test into "test" extras in pyproject.toml, and not rely on requirements-test anymore.
But I think that's out of the scope of this PR :)

.travis.yml
```diff
@@ -12,9 +12,12 @@ matrix:
...
-  - pip install -r requirements-test.txt
-  - pip install codecov
-script: py.test --cov=flit
+  - python3 -m pip install --upgrade pip
+  - python3 -m pip install --upgrade -r requirements-test.txt
+  - python3 -m flit install --deps develop
+  - python3 -m pip install codecov
+script:
+  - py.test --cov=flit
+  - sphinx-build -b html -d _build/doctrees doc _build/html
```
takluyver: If you're trying to verify that it builds without warnings, I think you need the `-W` flag to turn warnings into errors.

Also, could we do this inside an `if` so it only builds the docs on one version of Python?

takluyver:
I'm trying push flit forwards a bit. I've got the docs building without warnings now, with a slightly different approach.

If you're still interested in this (or anyone else reading this), I'd still be interested in having a convenient way to check that the docs build without warnings. Now it would make sense to integrate that into the tox config.

PR #202, opened 2018-08-29
Fix sdist.py

djmoch:
Address improper parsing of requirements.

Additionally, in cases where a script was specified in pyproject.toml,
the resulting setup.py would fail to parse the entry_points dictionary.
Fix by importing setuptools.


ignatenkobrain:
ping @takluyver

flit/sdist.py
```diff
@@ -89,11 +89,11 @@ def find_nearest_pkg(rel_path):
 def _parse_req(requires_dist):
     """Parse "Foo (v); python_version == '2.x'" from Requires-Dist
 
-    Returns pip-style appropriate for requirements.txt.
+    Returns requirements appropriate for setup.py
     """
-    if ';' in requires_dist:
+    if '; extra ==' in requires_dist:
```
takluyver: Sorry, that doesn't look right. Environment markers are more general than extras.

djmoch: I understand. What this if block seemed to be trying to do was separate the `install_requires` from the `extras_require`. Maybe I misunderstood the intent, but I did test this and it produced the desired results.

To test I built a source distribution of Flit itself. Here's a diff of the generated `setup.py` before and after my changes:
```
--- setup.py	1969-12-31 19:00:00.000000000 -0500
+++ setup.py.new	2018-09-07 12:47:45.061833303 -0400
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 # setup.py generated by flit for tools that don't yet use PEP 517
 
-from distutils.core import setup
+from setuptools import setup
 
 packages = \
 ['flit', 'flit.vcs', 'flit.vendorized', 'flit.vendorized.readme']
@@ -10,16 +10,19 @@
 {'': ['*'], 'flit': ['license_templates/*']}
 
 install_requires = \
-['requests', 'docutils', 'requests_download', 'pytoml']
+['requests',
+ 'docutils',
+ 'requests_download',
+ 'pytoml',
+ "zipfile36; python_version in '3.3 3.4 3.5'"]
 
 extras_require = \
-{':extra == "doc"': ['pygments-github-lexers'],
- ':extra == "test"': ['testpath',
-                      'responses',
-                      'pytest>=2.7.3',
-                      'pytest-warnings',
-                      'pytest-cov'],
- ":python_version in '3.3 3.4 3.5'": ['zipfile36']}
+{'doc': ['pygments-github-lexers'],
+ 'test': ['testpath',
+          'responses',
+          'pytest>=2.7.3',
+          'pytest-warnings',
+          'pytest-cov']}
 
 entry_points = \
 {'console_scripts': ['flit = flit:main']}
```

dholth: It's OK to just use extras_require (with an empty extra name for the install_requires) for everything. The ':extra == "doc"' in the extra name is unusual.

djmoch: It’s possible it would work. I can’t remember whether or not I tested a minimal change (just changing the `distutils` import to `setuptools`). I’m not opposed to backing out these changes and just leaving the import change if it works.

dholth: I agree with the +'s in this changeset.

takluyver:
Thanks - I wound up fixing this a different way as part of fixing a related issue in #212.

PR #203, opened 2018-08-30
Bump version to 1.2.dev0

Mariatta:


takluyver:
Thanks! I cut the `dev0` out of the docs again - no particular reason, that's just how I tend to do it.

PR #205, opened 2018-09-03
sdist: use setuptools

ignatenkobrain:
Unfortunately distutils doesn't support any of install_requires, python_require, etc…


djmoch:
See also, #202.

PR #206, opened 2018-09-05
Don’t check for toml file in “installfrom”

### LINKED ISSUE #158: Installing a Python package from Github fails
When I try to install a Flit package from Github, for example via

```
flit installfrom github:takluyver/flit
```

Flit only complains with the following message:

```
Neither pyproject.toml nor flit.ini found, and no other config file path specified
```

This also happens if I specify the full URL (https://github.com/takluyver/flit)

flying-sheep:
Fixes #158

### DIFF
--- a/flit/__init__.py
+++ b/flit/__init__.py
@@ -82,8 +82,11 @@ def main(argv=None):
     args = ap.parse_args(argv)
 
     cf = args.ini_file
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks! I like the readability improvement too.

PR #209, opened 2018-09-20
Uniform version constraint strings in docs

### LINKED ISSUE #208: Inconsistency in the docs about version constraints on pyproject.toml requires/requires-extra
As of 2018-09-19, the docs prescribe adding a version constraint in the `requires` section of `pyproject.toml` according to the `requests` line below:

```
requires = [
    "requests (>=2.6)",
    "configparser; python_version == '2.7'",
]
```

But for `requires-extra`, the docs say "The requirements are specified in the same format as for `requires`", but the example shows (see the pytest line)

```
test = [
    "pytest>=2.7.3",
    "pytest-cov",
]
``` 

`"requests (>=2.6)"` and `pytest>=2.7.3` are not in the same style.

I'd be happy to submit a PR when someone tells me what the fix should be (I'm guessing it should go all as in `requires`?).

gpfreitas:
Fixes #208.

Minor change to make version constraint strings in the `requires-extra` docs uniform/consistent.

### DIFF
--- a/doc/pyproject_toml.rst
+++ b/doc/pyproject_toml.rst
@@ -55,16 +55,16 @@ The remaining fields are optional:
 home-page
   A URL for the project, such as its Github repository.
 requires
-  A list of other packages from PyPI that this package needs. Each package
-  may be followed by a version specifier in
-  parentheses, like ``(>=4.1)``, and/or an `environment marker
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks :-)

PR #210, opened 2018-09-21
Fix #175 and add type annotation to arg. of `get_module_info`

### LINKED ISSUE #175: Mention filename when suggesting that user add a docstring
This one's more of a nicety.

Instead of the current message, when a docstring is missing, it would be nicer to have the path to the file suggested as well.

---

Suggesting changing:

> Cannot package module without docstring, or empty docstring. Please add a docstring to your module.

to something like:

> Cannot package module without docstring, or empty docstring. Please add a docstring to your module (/Users/pradyunsg/Projects/flit-trial/myscript.py).

gpfreitas:
- Add module name to error about missing docstring.
- Add type annotation to function common.get_module_info signature
- Adapt tests to the new error message.

### DIFF
--- a/flit/common.py
+++ b/flit/common.py
@@ -97,7 +97,7 @@ def get_docstring_and_version_via_import(target):
     version = m.__dict__.get('__version__', None)
     return docstring, version
 
-def get_info_from_module(target):
+def get_info_from_module(target: Module):
     """Load the module/package, get its docstring and __version__
     """
     log.debug("Loading module %s", target.file)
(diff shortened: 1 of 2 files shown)


gpfreitas:
In the test I check if the error message starts with the right leading sentence, then contains the filename. That was to keep the test code looking simple.

Let me know if you want a full equality check though, which may be "better".

gpfreitas:
Ok, I did the full equality checking, I think it's better and does not complicate the test much.

gpfreitas:
Forgot to say: I checked this is the only place where that error message occurs.

takluyver:
Thanks! I actually prefer tests checking error messages to be more loose: the error message isn't part of the API, and if I reword the error message, that's not a problem, so I don't want the tests to fail.

In this case, there's a specific `NoDocstringError` exception class. So I'd check for that, rather than the more generic ValueError, and then check that the message includes the filename.

takluyver:
Thanks :-)

PR #212, opened 2018-09-27
Fix using environment markers in requires-extra

### LINKED ISSUE #211: Requirement with extra and environment specifier is misconstructed
I got an error like this from PyPI:

```
requests.exceptions.HTTPError: 400 Client Error: Invalid value for requires_dist. Error: Invalid requirement: 'pathlib2; python_version == "2.7"; extra == "test"'. for url: https://upload.pypi.org/legacy/
```

The extra is being stapled on as an additional environment marker - it should be done with an `and` instead.

takluyver:
Closes #211. I've just successfully uploaded testpath 0.4.1 with this branch.

I shifted the code to translate `requires-extra` to `Requires-Dist` into `flit.inifile`, where I think it belongs.


merwok:
This seems great!  Any idea of a release date?

PR #220, opened 2018-10-12
implement: validate_environment_marker

### LINKED ISSUE #213: Validate parentheses properly in environment markers
According to [the grammar in PEP 508](https://www.python.org/dev/peps/pep-0508/#complete-grammar), environment markers can have parentheses, like this:

```
foo; extra == "test" and (os_name == "nt" or python_version == "2.7")
```

For now, I just strip leading/trailing parentheses from each clause before checking them. It would be good to check that the parentheses are balanced and only occur in the correct places, before and after individual boolean clauses.

I suspect that this requires a degree of 'proper' parsing of these expressions, rather than just regex-munging.

DreamN:
resolve #213

### DIFF
--- a/flit/validate.py
+++ b/flit/validate.py
@@ -215,6 +215,8 @@ def validate_environment_marker(em):
             else:
                 if idx > 1:
                     if MARKER_OP.match(stk[-1]):
+                        if stk[-2] == "EXP":
+                            raise Exception("Invalid expression")
                         l, op, r = token[idx - 2:idx + 1]
                         verify_statement(l, op, r)
                         stk.append(token[idx])
--- a/tests/test_validate.py
+++ b/tests/test_validate.py
@@ -82,11 +82,11 @@ def test_validate_environment_marker():
 
     res = vem("""()))))()extra == "test"(((((((""")  # No chained comparisons
     assert len(res) == 1
-    assert res[0] == 'Validation Error incorrect parentheses'
+    assert res[0] == 'Invalid expression. incorrect parentheses'
 
     res = vem("""extra == "test" and or (os_name == "nt" or python_version == "2.7")""")  # No chained comparisons
     assert len(res) == 1
-    assert res[0] == """<class 'list'>: ['Invalid expression "and or"']"""
+    assert res[0] == "Invalid expression \"and or\""
 
     assert len(vem('os.name == "linux\'')) == 2
 


flit/validate.py
```diff
@@ -149,37 +153,88 @@ def _valid_version_specifier(s):
...
+                        l, op, r = token[idx - 2:idx + 1]
+                        verify_statement(l, op, r)
+                        stk.append(token[idx])
+                        reduce_expression()
+                        if len(stk) > 1 and stk[-1] in {"and", "or"}:
+                            reduce_expression()
+                    else:
+                        stk.append(token[idx])
+                else:
+                    stk.append(token[idx])
```
takluyver: Can you add some comments in your parser code? I'm finding it a bit hard to follow.

flit/validate.py
```diff
@@ -149,37 +153,88 @@ def _valid_version_specifier(s):
...
+            idx += 1
+        if len(stk) != 1 or stk[-1] != "EXP":
+            problems.append("Invalid environment markers syntax")
+    except Exception as e:
```
takluyver: I don't like this structure - it could easily obscure a bug in the code by catching an overly general exception, and it means all the interesting code is indented a level more than it needs to be.

If you want to jump out of parsing as soon as a problem is detected, I think it's OK to use an early return.

tests/test_validate.py
```diff
@@ -76,6 +80,14 @@ def test_validate_environment_marker():
     assert len(res) == 1
     assert res[0].startswith("Invalid expression")
 
+    res = vem("""()))))()extra == "test"(((((((""")  # No chained comparisons
```
takluyver: The comment here is misleading, I think

tests/test_validate.py
```diff
@@ -76,6 +80,14 @@ def test_validate_environment_marker():
...
+    assert len(res) == 1
+    assert res[0] == 'Invalid expression. incorrect parentheses'
+
+    res = vem("""extra == "test" and or (os_name == "nt" or python_version == "2.7")""")  # No chained comparisons
+    assert len(res) == 1
+    assert res[0] == "Invalid expression \"and or\""
+
```
takluyver: Can we add  a test for an empty pair of parentheses, like `extra == "test" and ()`?

flit/validate.py
```diff
@@ -125,17 +127,19 @@ def _is_identifier_attr(s):
...
 VERSION_SPEC = re.compile(r'(~=|===?|!=|<=?|>=?)\s*[A-Z0-9\-_.*+!]+$', re.IGNORECASE)
 REQUIREMENT = re.compile(NAME.pattern[:-1] +  # Trim '$'
-     r"""\s*(?P<extras>\[.*\])?
+                         r"""\s*(?P<extras>\[.*\])?
          \s*(?P<version>[(=~<>!][^;]*)?
          \s*(?P<envmark>;.*)?
      $""", re.IGNORECASE | re.VERBOSE)
 MARKER_OP = re.compile(r'(~=|===?|!=|<=?|>=?|\s+in\s+|\s+not in\s+)')
 
+
```
takluyver: Not a big deal, but as a general rule, please try not to reformat code beyond the bit that you're working on.

flit/validate.py
```diff
@@ -125,17 +127,19 @@ def _is_identifier_attr(s):
...
     return problems
 
+
 # Distribution name, not quite the same as a Python identifier
 NAME = re.compile(r'^([A-Z0-9]|[A-Z0-9][A-Z0-9._-]*[A-Z0-9])$', re.IGNORECASE)
 r''
 VERSION_SPEC = re.compile(r'(~=|===?|!=|<=?|>=?)\s*[A-Z0-9\-_.*+!]+$', re.IGNORECASE)
 REQUIREMENT = re.compile(NAME.pattern[:-1] +  # Trim '$'
-     r"""\s*(?P<extras>\[.*\])?
+                         r"""\s*(?P<extras>\[.*\])?
```
takluyver: Please put the alignment here back how it was, it's like that deliberately.

DreamN:
I'll close this PR and use this PR #225  instead

PR #224, opened 2018-10-22
Support for Fossil VCS

takluyver:
From discussion on #133.

This code apparently worked for @richieadler, but I haven't checked it carefully yet, just added it to git. Anyone else with an interest in Fossil, please take a look!

### DIFF
--- a/flit/vcs/fsl.py
+++ b/flit/vcs/fsl.py
@@ -1,0 +1,32 @@
+import os
+from subprocess import check_output
+
+name = 'fossil'
+
+def find_repo_root(directory):
+    for p in [directory] + list(directory.parents):
+        if ((p / '.fslckout').is_file()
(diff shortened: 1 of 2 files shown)


flit/vcs/fsl.py
```diff
@@ -0,0 +1,32 @@
...
+name = 'fossil'
+
+def find_repo_root(directory):
+    for p in [directory] + list(directory.parents):
+        if ((p / '.fslckout').is_file()
+            or (p / '_FOSSIL_').is_file()):
+            return p
+
+def _repo_paths_to_directory_paths(paths, directory):
+    # 'fossil ls' gives paths from repo root, which may not be our directory.
```
richieadler, line 13: It may need to add the
```python
directory = directory.resolve()
```
that was included in `hg.py`, just to be consistent, but other than that, this still works.

PR #225, opened 2018-10-31
add validate parentheses environment markers

### LINKED ISSUE #213: Validate parentheses properly in environment markers

DreamN:
resolve #213

### DIFF
--- a/flit/validate.py
+++ b/flit/validate.py
@@ -134,7 +134,8 @@ REQUIREMENT = re.compile(NAME.pattern[:-1] +  # Trim '$'
          \s*(?P<version>[(=~<>!][^;]*)?
          \s*(?P<envmark>;.*)?
      $""", re.IGNORECASE | re.VERBOSE)
-MARKER_OP = re.compile(r'(~=|===?|!=|<=?|>=?|\s+in\s+|\s+not in\s+)')
+MARKER_OP_PATTERN = r'(?:~=|===?|!=|<=?|>=?|\s+in\s+|\s+not in\s+)'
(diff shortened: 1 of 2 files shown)


flit/validate.py
```diff
@@ -163,21 +164,37 @@ def validate_requires_python(metadata):
...
-        # TODO: validate parentheses properly. They're allowed by PEP 508.
-        parts = MARKER_OP.split(c.strip('()'))
-        if len(parts) != 3:
-            problems.append("Invalid expression in environment marker: {!r}".format(c))
+    marker_var_ops_value_pattern = r"""\b\S+\b """ + MARKER_OP_PATTERN + r""" (?:"\b\S+\b"|'\b\S+\b')"""
+
+    # check pattern "VAR OPS VALUE"
+    var_value_pair = re.findall(marker_var_ops_value_pattern, em)
+    for var_value in var_value_pair:
+        parts = re.compile(MARKER_OP_PATTERN).split(var_value)
```
earonesty: Use `MARKER_OP` instead of `re.compile(MARKER_OP_PATTERN)`

PR #226, opened 2018-11-02
py.test -> pytest

wimglenn:
py.test is not recommended since pytest v3.0 (see https://github.com/pytest-dev/pytest/issues/1629)


takluyver:
Thanks!

PR #229, opened 2018-11-03
Fix bogus flit init tests

wimglenn:
Previously, these were only passing because flit itself has a `pyproject.toml` and `LICENSE` file in project root.  If you assert on contents you would see for example email was actually `thomas@kluyver.me.uk` and not the expected `thomas@example.com`.


takluyver:
Good catch

PR #230, opened 2018-11-03
better validator for home-page field

wimglenn:
1. home-page is optional since https://github.com/takluyver/flit/issues/173 , so `flit init` shouldn't require that
2. `flit init` shouldn't accept a home-page value [for which `flit build` will refuse to build](https://github.com/takluyver/flit/blob/master/flit/validate.py#L210-L212)


takluyver:
Thanks!

PR #232, opened 2018-11-10
Check for Python 2 support a little more thoroughly

### LINKED ISSUE #227: requires-python parsing too picky
I had `requires-python = ">= 3.7"` and `flit build` made a wheel tagged `-py2.py3-none-any.whl`

If I put `requires-python = ">=3.7"` then I get the desired `-py3-none-any.whl`

zenbot:
Fixes #227.

### DIFF
--- a/tests/test_common.py
+++ b/tests/test_common.py
@@ -72,3 +72,21 @@ class ModuleTests(TestCase):
 def test_normalize_file_permissions():
     assert normalize_file_permissions(0o100664) == 0o100644 # regular file
     assert normalize_file_permissions(0o40775) == 0o40755   # directory
+
+@pytest.mark.parametrize(
+    ("requires_python", "expected_result"),
+    [
+        ("", True),
+        (">2.7", True),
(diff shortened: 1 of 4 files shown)


takluyver:
Thanks! This is a bit of code that's lingered from basically the first rough draft of Flit, and it's long overdue some improvement.

tests/test_common.py
```diff
@@ -72,3 +72,21 @@ def test_version_raise(self):
...
+
+@pytest.mark.parametrize(
+    ("requires_python", "expected_result"),
+    [
+        ("", True),
+        (">2.7", True),
+        ("3", False),
+        (">= 3.7", False),
+        ("<4, > 3.2", False),
+        ('>3.4', False),
```
takluyver: Can we have a test for the pattern people often use to indicate 2.7 or 3.x support, like `>=2.7, !=3.0.*, !=3.1.*, !=3.2.*`.

takluyver:
Thanks!

PR #233, opened 2018-11-13
Bugfix for building sdists in Mercurial repos

zenbot:
find_repo_root's parent-directory-walking behaviour doesn't behave quite as expected when called with a relative path; if called in any directory except the root of the repo it would return None, which calling code did not expect.


takluyver:
Seems reasonable, thanks!

PR #239, opened 2018-12-05
Allow specifying encodings

flying-sheep:
This e.g. allows using [future-fstrings](https://github.com/asottile/future-fstrings), which I’d like to use in flying-sheep/get_version#3

You can check that this works via:

```py
#!/usr/bin/env python3.5
import ast
st = ast.parse(b"# coding: future-fstrings\nprint(f'hi {ast}')")
exec(compile(st, filename="", mode="exec"))
# hi <module 'ast' from '/usr/lib64/python3.5/ast.py'>
```

### DIFF
--- a/flit/common.py
+++ b/flit/common.py
@@ -69,7 +69,8 @@ def get_docstring_and_version_via_ast(target):
     extracted by parsing its AST.
     """
     # read as bytes to enable custom encodings
-    node = ast.parse(target.file.read_bytes())
+    with target.file.open('rb') as f:
+        node = ast.parse(f.read())
     for child in node.body:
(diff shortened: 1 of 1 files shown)


takluyver:
The `Path.read_bytes()` method was new in 3.5, so there's a test failure on 3.4. However, I think we can move the minimum supported version to 3.5 at this point.

flying-sheep:
well, or we can open the file in binary mode. as you wish.

takluyver:
I'm OK with dropping 3.4, but I don't feel a strong need to, so since you've made it work with 3.4...

flying-sheep:
Great, thank you!

PR #240, opened 2018-12-17
Update filename in flit --help output

wimglenn:
It's still listing `flit.ini` but  that's not what's generated anymore.


takluyver:
Thanks, but can we lose the test? This wouldn't have caught the problem, because the test would have been out of date too, and testing a human-readable message just means it's slightly more effort whenever you update that message.

takluyver:
Thanks!

PR #254, opened 2019-03-13
Print `flit build` error msgs without exception traceback

carlbordum:


takluyver:
Thanks, seems reasonable

PR #263, opened 2019-05-01
Add "flit info --version" command

exhuma:
This PR adds a new subcommand `info` with the argument `--version`.

Calling `flit info --version` will simply print the package version number to stdout. As the version number is missing in `pyproject.toml` when using `flit`, this command allows external tools to easily access the projects version number.

I did not implement this as `flit --version` as that woulc conflict with the existing flag. Having is as subcommand also makes it a bit more expressive. This PR adds the new subcommand "info" which could serve as a container for additional "informational" commands.

Questions to the maintainer(s):

I also - out of habit - added a type annotation. To be on the safe side, I used the Python2 syntax. But personally I prefer the Python 3 syntax. What is your position on this? If it is OK with you I will switch to the Py3 syntax.

Unrelated to this PR, I will also quickly refactor the subcommands. This will be a fairly big change and might trigger some discussion. How do hold discussions? Do you have a mailing-list? A IRC channel? Or should I simply add an issue & PR in "WIP" status and use GitHub comments for discussion?

### DIFF
--- a/flit/__init__.py
+++ b/flit/__init__.py
@@ -79,6 +79,14 @@ def main(argv=None):
         help="Prepare pyproject.toml for a new package"
     )
 
+    parser_info = subparsers.add_parser('info',
+        help="Retrieve metadata information from the project",
+    )
+    parser_info.add_argument(
+        '--version', default=False, action='store_true', dest='show_version',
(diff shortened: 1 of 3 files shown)


earonesty:
Looks good to me.   Simple addition.

PR #264, opened 2019-05-01
Refactor subcommands

exhuma:
As mentioned in #263, this is a refactoring for subcommands.

Let me know what you think. It builds on top of my code from #263 so this also includes the commit of the `info --version` command. I kept them separate in case this refactoring is not something you like in terms of code-style.


earonesty:
nicely organized, and lots of extra tests

PR #268, opened 2019-05-15
Add missing doc dependencies in pyproject.toml

cdeil:
I had to add these to make the docs build work.

These instructions don't give a dev env where the docs can be generated locally:
https://github.com/takluyver/flit#development

I found this option:
```
flit install --deps all
```
Maybe recomment that to hack on `flit`?


takluyver:
Thanks!

PR #270, opened 2019-05-15
Add docs/packages.rst

cdeil:
@takluyver - What do you think?


cdeil:
@takluyver - Thoughts?

PR #272, opened 2019-05-30
add enscons build

dholth:
Here's an enscons SConstruct for you to enjoy. :) Install enscons from pypi, and just run scons to build the sdist and wheel. I've given the wheel an invalid tag so it doesn't clobber the flit generated one.

I did notice something that might be useful to you: flit didn't deflate all the files in its own wheel.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,6 @@ __pycache__/
 /.coverage
 /.pytest_cache
 /.tox
+*.dblite
+*.egg-info
+PKG-INFO
--- a/SConstruct
+++ b/SConstruct
@@ -1,0 +1,53 @@
+import os
+import enscons
+import pytoml as toml
+
+from subprocess import check_output
+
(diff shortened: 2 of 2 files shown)


takluyver:
Thanks! I'm not going to merge this, as Flit builds itself. You probably saw that coming. But it's interesting to see what it looks like.

And thanks for the tip about deflating files; I'll have a look to see if that's still the case now.

takluyver:
Ha, turns out I've been overlooking compression on most of the files in every wheel, probably for several years. The wheel for flit is less than half the size once you turn on compression.

Maybe people shouldn't use software I write. :wink:

takluyver:
Commit 7f1efcdc8ff1d1298846b5552999e9bdebfe9061 enables compression - thanks for spotting that oversight.

I'll do another release soon, but maybe I'll leave it until tomorrow in case more problems come to light.

PR #278, opened 2019-07-26
Fix CI error on test_misspelled_key

Carreau:
Error seem to be due to pytest wrapping errors, and the str()
representation of the error not always having the full error message of
the underlying error.

### DIFF
--- a/tests/test_inifile.py
+++ b/tests/test_inifile.py
@@ -40,7 +40,7 @@ def test_misspelled_key():
     with pytest.raises(ConfigError) as e_info:
         read_pkg_ini(samples_dir / 'misspelled-key.ini')
 
-    assert 'description-file' in str(e_info)
+    assert 'scription-file' in str(e_info.value)
 
 def test_description_file():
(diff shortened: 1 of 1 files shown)


Carreau:
I have issues reproducing the error locally so this is kind-of a shot in the dark.

takluyver:
The [pytest 5.0.0 release notes](https://docs.pytest.org/en/latest/changelog.html#pytest-5-0-0-2019-06-28) mention that you should change `assert "some message" in str(e)` to `assert "some message" in str(e.value)`. Does this work without changing the string?

takluyver:
Yup, that seems to be OK.

Carreau:
> The [pytest 5.0.0 release notes](https://docs.pytest.org/en/latest/changelog.html#pytest-5-0-0-2019-06-28) mention that you should change `assert "some message" in str(e)` to `assert "some message" in str(e.value)`

Ah thanks I missed that. 

Hope you had some good time off !

takluyver:
I did, thanks :-)

PR #282, opened 2019-08-01
WIP: Separate flit_core

takluyver:
This is still in flux, but it's showing the direction which I'm planning to go in for Flit 2.0.

Flit 0.x & 1.x were written as only developer tools, creating packages which downstream consumers would get without any further need for Flit. But PEP 517 has a different model, in which you can request a from-source install, and Flit will run as a build tool on the target Python version. This only works if the build tool works on that Python version, so we've gone backwards for installing on Python 2 (it still works normally, but not if you request installation from source).

So this separates out `flit_core`, a PEP 517 backend, from `flit`, the surrounding developer tools. The idea is that flit_core will be conservative with supported Python versions and dependencies, perhaps even supporting Python 2 (I haven't fully decided yet), while flit can be more aggressive. My plan for now is that they will live in the same git repository, and be released in lockstep, but exist as two separate packages on PyPI.

`flit_core` currently builds a minimal sdist for PEP 517, whereas `flit build` extends this to make an 'archival' sdist with the files checked in to version control, and an optional `setup.py` for backwards compatibility. I'm also toying with the idea of dropping the requirement for version control, and/or allowing pyproject.toml to include & exclude files for the sdist, but these are not implemented yet.


okken:
At least an option to build without version control would be cool. Or switch default to allow no version control and a flag to enforce VC. 

Also I would personally vote for core to be a separate repo if it’s a separate install. But in any case, the work to separate core sounds like a cool idea.

takluyver:
Thanks Brian!

I'm wary of splitting up the repo from my experience with Jupyter & IPython: it's much harder for people to find existing issues, so you get much more confusion and duplication. So for now, I'm aiming for a one project, two packages approach. I might change that in the future.

PR #286, opened 2019-08-18
fix shebang in scripts installed with --symlink 

### LINKED ISSUE #285: flit install --symlink doest not honor --python for scripts it installs
Hi,

I have been experimenting with flit recently and I very much like it's approach for development installs (`--symlink`).

I noticed what could be a small issues with such installations.

I have flit installed globally (with pipx).

When I try to install my project with `pip install --symlink --python $(which python)` in a virtualenv, I notice the installed script uses flit's python as shebang instead of the python of the virtualenv.

I believe the problem arises in this `install_script`, called via `install_directly`:
https://github.com/takluyver/flit/blob/f5996dcee05656ca7a1e6ed2e5c82a6e911f4f10/flit/install.py#L171

sbidoul:
Fixes #285 

I also fixed a deprecation warning for an invalid escape sequence.

### DIFF
--- a/tests/test_install.py
+++ b/tests/test_install.py
@@ -107,6 +107,8 @@ class InstallTests(TestCase):
         assert_islink(self.tmpdir / 'site-packages2' / 'package1',
                       to=str(samples_dir / 'package1'))
         assert_isfile(self.tmpdir / 'scripts2' / 'pkg_script')
+        with (self.tmpdir / 'scripts2' / 'pkg_script').open() as f:
+            assert f.readline().strip() == "#!mock_python"
 
     def test_install_requires(self):
         ins = Installer(samples_dir / 'requires-requests.toml',
--- a/tests/test_install.py
+++ b/tests/test_install.py
@@ -37,6 +37,8 @@ class InstallTests(TestCase):
         assert_isdir(self.tmpdir / 'site-packages' / 'package1')
         assert_isdir(self.tmpdir / 'site-packages' / 'package1-0.1.dist-info')
         assert_isfile(self.tmpdir / 'scripts' / 'pkg_script')
+        with (self.tmpdir / 'scripts' / 'pkg_script').open() as f:
+            assert f.readline().strip() == "#!" + sys.executable
 
     def test_symlink_package(self):
         if os.name == 'nt':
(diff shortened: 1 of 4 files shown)


tests/test_install.py
```diff
@@ -107,6 +111,8 @@ def test_symlink_other_python(self):
         assert_islink(self.tmpdir / 'site-packages2' / 'package1',
                       to=str(samples_dir / 'package1'))
         assert_isfile(self.tmpdir / 'scripts2' / 'pkg_script')
+        with (self.tmpdir / 'scripts2' / 'pkg_script').open() as f:
+            assert f.readline().strip() == "#!mock_python"
```
takluyver: This makes me think we should probably do an `abspath` on it. I don't think we want shebangs with a relative path.

sbidoul: That should probably go very early in the code, just after argument parsing.
That probably involves searching in `PATH` too, because `flit install --python python` just works today as way to install in the current active virtualenv.

sbidoul: A reliable (although a bit inefficient) possibility could be something like this

```python
def _normalize_python(python):
    if python == sys.executable:
        return python
    return subprocess.check_output([python, "-c", "import sys; print(sys.executable)"]).strip()
```

sbidoul: possibly with a shortcut if `isabs(python)`

sbidoul: In any case, one could argue that flit works better with this PR than not, if only for people providing a full path in `--python`. So maybe it could be merged as is? Alternatively, if you think the above idea about normalizing `args.python` is valid, I'm happy to add a commit with it to this PR.

takluyver: Yes, good points. In most cases I think `os.path.abspath(shutil.which(self.python))` should give the right answer, but it wouldn't surprise me if there's some odd corner cases with symlinks and things where that's different from running it in a subprocess and getting `sys.executable`.

I agree that this PR is a strict improvement, so I'm happy to merge this as if if you prefer.

sbidoul:
Interestingly, one can see in the last commit that shutil.which and subprocess seem to produce different results.

image

sbidoul:
Do I amended the last test. This should be complete now :)

takluyver:
Thanks. That difference is strange - it looks like `shutil.which()` is getting the wrong path somehow.

sbidoul:
Now if only pip uninstall would support those symlinks..

sbidoul:
> Now if only pip uninstall would support those symlinks..

Updat:e: pip >= 19.3 now uninstalls symlinks. (https://github.com/pypa/pip/pull/6914)

PR #288, opened 2019-10-15
Add support for Python 3.8

hugovk:
Released yesterday: https://discuss.python.org/t/python-3-8-0-is-now-available/2478

We can also simplify the CI matrix a bit because Xenial is now default.


takluyver:
:+1: thanks again.

I'l try to get round to finishing up #282, which will lead to Flit 2.

PR #289, opened 2019-10-24
make Author-email field optional

wimglenn:
The core metadata spec says it should be optional  https://packaging.python.org/specifications/core-metadata/#author-email

Related: https://github.com/takluyver/flit/issues/236

Anecdotally, I've seen people confused about what to do when they don't want a public email associated (currently they can only use flit by entering a fake email address there)

### DIFF
--- a/flit/common.py
+++ b/flit/common.py
@@ -234,7 +234,7 @@ class Metadata:
     def __init__(self, data):
         self.name = data.pop('name')
         self.version = data.pop('version')
-        self.author_email = data.pop('author_email')
+        self.author_email = data.pop('author_email', None)
         self.summary = data.pop('summary')
 
         for k, v in data.items():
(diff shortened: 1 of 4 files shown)


okken:
I'd love to see this merged.
I currently put in a fake email during init and replace it with an empty string later.

flit/common.py
```diff
@@ -234,7 +234,7 @@ class Metadata:
     def __init__(self, data):
         self.name = data.pop('name')
         self.version = data.pop('version')
-        self.author_email = data.pop('author_email')
+        self.author_email = data.pop('author_email', None)
```
takluyver: Let's handle this like the other optional fields - remove this line, and set `author_email = None` on the class.

takluyver:
Thanks, this looks OK apart from one minor change. Unfortunately my recent reorganisation has caused a merge conflict that github can't resolve. Do you want to resolve that? I'm happy to fix it, as it was me who broke it.

wimglenn:
Hi Thomas, I’ll leave it in your capable hands. Cheers!

PR #295, opened 2019-11-23
Add environ FLIT_INSTALL_PYTHON

### LINKED ISSUE #294: Config to let flit install default to active virtual environment
Use case: I install only one copy of `flit` (user-wide), and use it for all projects I work on. But even if I activate the environment, `flit install` would prefer the Python it’s installed under over the environment the shell is currently in. My current “workaround” is to use `flit install --python=python`, but that seems quite redundant (and I sometimes forget it with frustrating results).

Proposed solution: Let Flit detect the `VIRTUAL_ENV` environment variable in `find_python_executable`, and use it to choose the executable if avaiable. This is backwards-incompatible, and would need to be behind a config (environ?). So maybe something like…

```py
def _find_default_python_executable():
    if not os.environ.get('FLIT_INSTALL_USE_VIRTUALENV'):
        return sys.executable
    prefix = os.environ.get("VIRTUAL_ENV')
    if not prefix:
        return sys.executable
    if sys.platform == 'win32':
        binpath = os.path.join(prefix, 'Scripts', 'python.exe')
    else:
        binpath = os.path.join(prefix, 'bin', 'python')
    if os.path.isfile(binpath):
        return binpath
    return sys.executable

def find_python_executable(python):
    if not python:
        return _find_default_python_executable()
    ...  # unchanged
```

uranusjr:
Close #294.

What should I supply to `.. versionadded::` in the documentation?

### DIFF
--- a/doc/cmdline.rst
+++ b/doc/cmdline.rst
@@ -172,6 +172,12 @@ Environment variables
    rejects your valid metadata. If you need to use it and you believe your
    metadata is valid, please `open an issue <https://github.com/takluyver/flit/issues>`__.
 
+.. envvar:: FLIT_INSTALL_PYTHON
+
+   Set a default Python interpreter for ``flit install`` to use when
+   ``--python`` is not specified. Flit uses the same environment itself is
+   installed in if this is left unset or empty.
+
 .. envvar:: SOURCE_DATE_EPOCH
 
    To make reproducible builds, set this to a timestamp as a number of seconds
--- a/doc/cmdline.rst
+++ b/doc/cmdline.rst
@@ -107,9 +107,12 @@ Install the package on your system.
 .. option:: --python <path to python>
 
    Install for another Python, identified by the path of the python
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks. Let's call it 2.1, as this is a new feature.

I'd mention in the docs that it can be either an absolute path, or a command name which will be found in PATH.

uranusjr:
Documentation section extended. I also added a cross-ref in the `--python` option part.

takluyver:
Thanks!

Ricyteach:
It took me quite a while to figure out that I can just supply "python" as the env variable value to have flit install to the current virtual environment. I feel this capability could be made much more clear in the docs.

Can the docs be updated to be a little bit clearer about what "command name" means? Perhaps by adding this specific "FLIT_INSTALL_PYTHON = python" example, since it's probably a use case that will come up fairly often?

Issue #222, opened 2018-10-17
system-wide flit does not install scripts in a virtualenv


mehdisadeghi:
`flit install` yields different behaviors when `flit` is installed in virtualenv compared to a system-wide installation.

If I use system-wide `flit` in a virtualenv it does not install the scripts defined in `[tool.flit.scripts]` section of `pyproject.toml` file without reporting any errors. However, if I install `flit` in the virtualenv and then run `flit install` the script will be installed in the virtualenv.

This is my `pyproject.toml`:
```
[build-system]
requires = ['flit']
build-backend = 'flit.buildapi'

[tool.flit.metadata]
module = 'mymodule'
requires-python = '>=3.5'
description-file = 'README.md'

[tool.flit.scripts]
myscript = 'mymodule.cli:cli'
```

I am using Flit 1.2.1 on macOS Mojave.

mehdisadeghi:
btw, I ran `flit build` and then installed the wheel with `pip` inside the virtualenv and pip installed the script correctly.

mehdisadeghi:
For future visitors: `flit install --env` gets the job done.

takluyver:
I'm pretty sure that `--env` will still install into the same environment that Flit itself is installed into, unless you use it in combination with `--python path/to/env/bin/python`.

`--env` is designed to be the counterpart to the `--user` flag. Flit tries to autodetect whether it should do a user install (like `pip install --user`), and those two flags let you force the answer to yes or no.

mehdisadeghi:
@takluyver my bad. You are absolutely right. Apparently I was pretty confused due to making too many changes at once. Thanks for the update.

jonathanchukinas:
Can we add this to [the docs](https://flit.readthedocs.io/en/latest/cmdline.html#cmdoption-flit-install-env)? I definitely misunderstood the --env option just like @mehdisadeghi until I combed through the issues and found this thread.

takluyver:
Sure, let's try to clarify it. Do you want to make a PR? I might reword it - I can get a bit particular about docs - but if you suggest what wording would have made it clearer for you, we can at least use that as a reference point.

jonathanchukinas:
Ok, perhaps something like this:

Under [Flit Install](https://flit.readthedocs.io/en/latest/cmdline.html#flit-install), we'd see:

> Install the package on your system. By default, the install location matches the flit installation (virtual environment vs. user-local). So if you run a flit installed in a virtual environment, the package will be installed there too. 
>
[...]

PR #299, opened 2019-11-24
Try to clarify how to control where packages are installed

takluyver:
From discussion on #222

@jonathanchukinas: would this be clear from your perspective?

### DIFF
--- a/doc/cmdline.rst
+++ b/doc/cmdline.rst
@@ -70,6 +70,15 @@ or another repository.
 
 Install the package on your system.
 
+By default, the package is installed to the same Python environment that Flit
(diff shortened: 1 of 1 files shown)


jonathanchukinas:
It's very close. Maybe just one more addition? Mention that flit has to be installed in that environment? See, I thought I was meeting the criteria of "being in a virtual env" by having activated that env without having flit installed into that env. To me, that seemed a reasonable interpretation of what I was reading. It would have helped me to have that explicitly called out.

takluyver:
Is this sentence not explicit enough?

> By default, the package is installed to the same Python environment that Flit itself is installed in

takluyver:
I'm going to merge this, as I think it's an improvement, and we can always make more improvements later.

PR #300, opened 2019-11-27
Make Python command absolute before calling subprocess

uranusjr:
[bpo-38905](https://bugs.python.org/issue38905)

On Python 3.7.2 or later, `sys.executable` returns the wrong executable inside venv on Windows unless you invoke the interpreter with an absolute path. It seems that the behaviour is expected and unlikely to be fixed (at least not any time soon). Core devs seem to approve the `shutil.which()` workaround.

### DIFF
--- a/flit/__init__.py
+++ b/flit/__init__.py
@@ -16,6 +16,8 @@ __version__ = '2.1.0'
 log = logging.getLogger(__name__)
 
 
+class PythonNotFoundError(Exception): pass
+
 def find_python_executable(python):
     if not python:
         python = os.environ.get("FLIT_INSTALL_PYTHON")
@@ -23,7 +25,9 @@ def find_python_executable(python):
         return sys.executable
     if os.path.isabs(python):  # sys.executable is absolute too
         return python
-    python = shutil.which(python) or os.path.abspath(python)
+    python = shutil.which(python)
+    if not python:
+        raise PythonNotFoundError('Python executable {!r} not found'.format(python))
     return subprocess.check_output(
         [python, "-c", "import sys; print(sys.executable)"],
         universal_newlines=True,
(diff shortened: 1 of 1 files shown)


takluyver:
That's rather frustrating. I've worked on a bunch of code that relies on `sys.executable` giving me an absolute, reliable path, regardless of how Python was invoked.

If we're using `which()` to get an absolute path anyway, does then finding `sys.executable` ever make any difference?

takluyver:
I was just about to post a grumpy message to that Python issue, but on reading it more carefully, I think I see it's not `sys.executable` that is giving you the wrong result, but the subprocess launch that's doing something different from a normal PATH search. That's less concerning; I'm OK with doing a `which()` on it in Python.

flit/__init__.py
```diff
@@ -22,6 +23,7 @@ def find_python_executable(python):
         return sys.executable
     if os.path.isabs(python):  # sys.executable is absolute too
         return python
+    python = shutil.which(python) or os.path.abspath(python)
```
takluyver: Why the `abspath` call? I think `which()` already handles relative paths.

uranusjr: I was not aware of this, thanks! TIL 😄

uranusjr: What should we fall back to if the `which` call fails though? (It returns `None` if there’s no matching executable found.) Just the unresolved `python`?

takluyver: We should probably throw an error, I'd guess - that means you've given it a name which isn't on PATH, or a path which doesn't exist or isn't executable.

uranusjr: I added an exception for this.

takluyver:
Thanks, this LGTM.

PR #303, opened 2019-12-16
Correctly set up package_dir for src/ prefix

### LINKED ISSUE #302: Generated setup.py in sdist does not configure package_dir correctly for src/ setup
From uranusjr/pyem#4.

The generated setup.py needs a `package_dir=""': "src"}` line for setuptools to correctly discover packages in `src/`.

uranusjr:
This adds something like

```python
package_dir = \
{'': 'src'}

setup(...
      package_dir=package_dir,
)
```

to setup.py if the module is found inside a `src/` prefix.

Fix #302.

Syntax reference: https://docs.python.org/3.9/distutils/setupscript.html#listing-whole-packages

### DIFF
--- a/flit_core/flit_core/common.py
+++ b/flit_core/flit_core/common.py
@@ -28,18 +28,22 @@ class Module(object):
         if osp.isdir(pkg_dir):
             self.path = pkg_dir
             self.is_package = True
+            self.prefix = ''
             existing.add(pkg_dir)
         elif osp.isfile(py_file):
             self.path = py_file
             self.is_package = False
+            self.prefix = ''
             existing.add(py_file)
         elif osp.isdir(src_pkg_dir):
             self.path = src_pkg_dir
             self.is_package = True
+            self.prefix = 'src'
             existing.add(src_pkg_dir)
         elif osp.isfile(src_py_file):
             self.path = src_py_file
             self.is_package = False
+            self.prefix = 'src'
             existing.add(src_py_file)
         else:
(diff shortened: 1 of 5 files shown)


takluyver:
Looks reasonable, but can we have a test? The `test_sdist` module already has a handy function to get global variables from a generated `setup.py`, so it should be easy to add.

flit_core/flit_core/common.py
```diff
@@ -28,18 +28,22 @@ def __init__(self, name, directory='.'):
         if osp.isdir(pkg_dir):
             self.path = pkg_dir
             self.is_package = True
+            self.prefix = ''
             existing.add(pkg_dir)
         elif osp.isfile(py_file):
```
takluyver: I just noticed that these `elif` should be `if` - we're checking that only one of the possible paths exists, but `elif` assumes that's the case and bypasses the check.

It's not related to this change, but if you're feeling keen, another PR to add a failing test and fix that would be welcome. If not, I'll get round to it at some point.

uranusjr:
My work PC just broke down :( I’m sending it for replacement tomorrow and will come back to this when I can (hopefully not for too long).

takluyver:
No problem, thanks for letting me know.

flit_core/flit_core/common.py
uranusjr: Does this mean that we want to implement additional checks to make sure the choices are mutually exclusive?

PR #306, opened 2019-12-27
Make bootstrap_dev work on Python 3.5

uranusjr:
Since Flit itself support down to 3.5, the bootstrap script probably needs to as well. Unfortunately pathlib support was spotty back then.


takluyver:
Thanks!

PR #307, opened 2020-01-04
Fix detection of conflicting module files

takluyver:
The code to check for this was already there, but because we were using `elif` in place of `if`, it could never actually fire.

Fixing this revealed that there was a conflict in the test sample files, which also needed to be cleaned up.


takluyver:
Thanks for checking it, @uranusjr :-)

PR #308, opened 2020-01-06
Fix module build logic to test all variants

uranusjr:
Finish up the work mentioned in https://github.com/takluyver/flit/pull/303#discussion_r358793408


uranusjr:
Heh so this actually catches a bug in tests!

takluyver:
Heh, snap: #307. Thanks for following this up, but I might go with my branch, as I already fixed the test failure and added a test for the behaviour I wanted to fix. Want to have a glance at that before I merge it, as you're thinking about the same thing?

uranusjr:
Happy to!

Issue #304, opened 2019-12-17
Option to disable generated setup.py in sdist


takluyver:
In https://github.com/uranusjr/pyem/issues/4 , @uranusjr said:

> FWIW I don’t even want to generate that setup.py; Flit generates it for me, and there’s no way to turn it off.

The setup.py was always intended as a temporary measure, and eventually some version of Flit will stop producing it. I don't think we're there yet, but I think it would now be reasonable to let package authors disable it if they prefer to go all in on PEP 517. That will give us some chance to figure out any problems before it becomes the default.

takluyver:
Anyone who's interested in this, take a look at PR #311, which adds a `--no-setup-py` option to the relevant commands.

PR #311, opened 2020-01-13
Add an option to make sdists without setup.py

takluyver:
The setup.py is a compatibility measure until we can rely on PEP 517. This lets people who want to bring the future forwards start relying on PEP 517 now.

Closes gh-304


takluyver:
Released as part of Flit 2.2.

PR #312, opened 2020-01-16
Allow for a LICENSES directory when writing wheel metadata

davidfritzsche:
The paths matched by `COPYING*` and `LICENSE*` when adding license files
to the dist-info directory are now allowed to be directories
containing the actual license files.

### DIFF
--- a/flit_core/flit_core/wheel.py
+++ b/flit_core/flit_core/wheel.py
@@ -195,7 +195,12 @@ class WheelBuilder:
 
         for base in ('COPYING', 'LICENSE'):
             for path in sorted(glob(osp.join(self.directory, base + '*'))):
-                self._add_file(path, '%s/%s' % (self.dist_info, osp.basename(path)))
+                if osp.isdir(path):
+                    for full_path in sorted(glob(osp.join(path, '*'))):
+                        rel_path = osp.relpath(full_path, self.directory)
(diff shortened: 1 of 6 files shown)


takluyver:
Thanks, this seems reasonable. But I'd like to think a bit more about what the limits are on this feature.

1. This allows for the top level items (`LICENSE*` or `COPYING*`) to be files or directories. But if they're directories, it assumes that the second level items are only files. That seems reasonable enough, but do we have any evidence that subdirectories are rare? What does the failure look like if there are subdirectories?
2. Should the names for directories be more constrained than the names we match for files? I think the wildcard was primarily so we pick up different extensions (`.txt` or `.md` or `.rst`), which doesn't matter so much for directories. Can we get any systematic look at what people call directories containing license files? Especially as the glob matching might not be case sensitive on all platforms, it shouldn't pick up unexpected directories.

I'm not aware of any tools that actually consume license information from wheels or installed packages, nor of any standard for how they should be stored. So if this is to be useful, it might be worth having a broader discussion on e.g. https://discuss.python.org/ or the [distutils-sig mailing list](https://mail.python.org/mailman3/lists/distutils-sig.python.org/).

PhilThaReal (approved):

AA-Turner:

A

Issue #273, opened 2019-07-02
Why flit?


guettli:
I look at this page: https://flit.readthedocs.io/en/latest/index.html

I have never used flit before and I don't know why I should do it.

Please create  a new section before "Install". Maybe with the title "why" or "Introduction".

Please tell us why you think flit should get used.

madsmtm:
There's a bit of motivation in the [reproducible builds](https://flit.readthedocs.io/en/latest/reproducible.html) section, but agreed, the documentation could use a dedicated "why" section

takluyver:
Thank you for the suggestion, but for now I'm happy with the current state of the landing page.

guettli:
Now I (and other people) understand soon what this package provides. Thank you!

brainwane:
@takluyver I was trying to understand/remember/tell others, just now what `flit`'s design philosophy is and what it tries to provide users. It wasn't obvious to me from the landing page of the docs. I dug around and found #283:

> one alternative to `setuptools` for certain use cases (the simple ones).

and https://github.com/takluyver/flit/issues/20#issuecomment-93020675 :

> I don't want to make `flit init` too complex - it's supposed to be a quick way to get a simple, working config file (with license choice integrated to encourage people to give their code a license). Let's focus on building and uploading packages for now. At some point, perhaps a nicer interactive package metadata editor would be good as a separate tool.

and this issue and the associated pull requests, which helped me a lot.

Is there some way we could include some information about those points somewhere a bit more prominently in the documentation? Or in [the Python Packaging User Guide list entry on flit](https://packaging.python.org/key_projects/#flit) ([source](https://github.com/pypa/packaging.python.org/blob/master/source/key_projects.rst#flit))?

takluyver:
OK, taking another look.

takluyver:
I've drafted some wording in #314.

brainwane:
Thank you @takluyver !

PR #314, opened 2020-02-02
Add section about why you might want Flit, and what other options exist

takluyver:
From #273.

### DIFF
--- a/doc/rationale.rst
+++ b/doc/rationale.rst
@@ -1,0 +1,58 @@
+Why use Flit?
+=============
+
+*Make the easy things easy and the hard things possible* is an old motto from
+the Perl community. Flit is entirely focused on the *easy things* part of that,
+and leaves the hard things up to other tools.
+
+Specifically, the easy things are pure Python packages with no build steps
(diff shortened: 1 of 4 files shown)


doc/rationale.rst
```diff
@@ -0,0 +1,58 @@
...
+like :pep:`518` and :pep:`517`, which are now used by other packaging tools
+such as `Poetry <https://python-poetry.org/>`_ and
+`Enscons <https://pypi.org/project/enscons/>`_.
+
+Other options
+-------------
+
+If your package needs a build step, you won't be able to use Flit.
+`Setuptools <https://setuptools.readthedocs.io/en/latest/>`_ is the de-facto
+standard, but newer tools such as Enscons_ also cover this case.
```
henryiii: How about mentioning scikit-build here? It's the same sort of idea as Enscons, but wraps CMake instead, and is also from KitWare.

takluyver: Thanks, that's interesting to know about, but I don't want to try to make this a comprehensive list, and scikit-build looks like more an extension of setuptools than a totally separate build tool. So I'd prefer to leave this for now.

PR #315, opened 2020-02-03
Replace 'pytoml' with 'toml'.

kalekundert:
The pytoml package is deprecated, and doesn't support the most recent TOML standard (e.g. dotted keys).

### DIFF
--- a/doc/development.rst
+++ b/doc/development.rst
@@ -5,7 +5,7 @@ To get a development installation of Flit itself::
 
     git clone https://github.com/takluyver/flit.git
     cd flit
-    python3 -m pip install docutils requests pytoml
+    python3 -m pip install docutils requests toml
     python3 bootstrap_dev.py
 
 This links Flit into the current Python environment, so you can make changes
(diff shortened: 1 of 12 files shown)


takluyver:
pip would also need to change before you could use new toml features, and it was suggested that pip might use the `tomlkit` library - see https://github.com/pypa/pip/issues/6120 . I'd like to follow pip once it jumps one way or another.

kalekundert:
I see, that makes sense.

jbrobst:
@takluyver Could this be merged now that pip has [switched to `toml`](https://github.com/pypa/pip/pull/8045)? `toml` is being [maintained](https://github.com/uiri/toml/commits/master) and there is still [interest in this issue](https://github.com/takluyver/flit/issues/255).

takluyver:
Makes sense, yup.

takluyver:
This PR now has merge conflicts. @jbrobst, as you're interested, would you like to make a new PR?

takluyver:
Merged as #378. Thanks both!

PR #319, opened 2020-02-20
ensures cache dir does exist

holgi:
I ran into a problem when installing another project with flit for development purpose using `flit install --pth-file`.  The user had a non-existing home directory set (`/nonexistent') and therefore the download of the trove classifiers failed.

The `/nonexistent` "home" directory is used on POSIX systems to denote a user without a home directory, e.g. user `nobody` or users only set up for services.

These proposed changes ensures that a valid directory is returned by the `flit.validate.get_cache_dir()` function. 

If the `flit`-subdirectory of the cache-directory does not exist and could not be created, a temporary directory is used instead. To ensure that always the same temporary directory is used, the function is wrapped with the `functors.lru_cache()`.

A tests for this new behavior is also included. The tests passed so far on Mac OS X (Python 3.7 and 2.7) and on FreeBSD (Python 3.6)

### DIFF
--- a/flit/validate.py
+++ b/flit/validate.py
@@ -1,6 +1,5 @@
 """Validate various pieces of packaging data"""
 
-import contextlib
 import io
 import logging
 import os
@@ -8,62 +7,72 @@ from pathlib import Path
 import re
 import requests
 import sys
-import tempfile
 
 from .vendorized.readme.rst import render
 
 log = logging.getLogger(__name__)
 
-@contextlib.contextmanager
 def get_cache_dir() -> Path:
     """Locate a platform-appropriate cache directory for flit to use
 
-    Ensures that the cache directory exists.
+    Does not ensure that the cache directory exists.
     """
     # Linux, Unix, AIX, etc.
     if os.name == 'posix' and sys.platform != 'darwin':
         # use ~/.cache if empty OR not set
-        local = os.environ.get("XDG_CACHE_HOME", None) \
+        xdg = os.environ.get("XDG_CACHE_HOME", None) \
               or os.path.expanduser('~/.cache')
+        return Path(xdg, 'flit')
 
     # Mac OS
     elif sys.platform == 'darwin':
-        local = os.path.expanduser('~/Library/Caches')
+        return Path(os.path.expanduser('~'), 'Library/Caches/flit')
 
     # Windows (hopefully)
     else:
         local = os.environ.get('LOCALAPPDATA', None) \
                 or os.path.expanduser('~\\AppData\\Local')
+        return Path(local, 'flit')
 
-    cache_dir = Path(local, "flit")
-    try:
-        cache_dir.mkdir(parents=True)
-    except FileExistsError:
-        pass
-    except (OSError, PermissionError):
-        temp_dir =  tempfile.TemporaryDirectory()
-        cache_dir = Path(temp_dir.name)
-    yield cache_dir
 
-def _verify_classifiers_cached(classifiers, cache_dir):
+def _read_classifiers_cached():
     """Check classifiers against the downloaded list of known classifiers"""
-    with (cache_dir / 'classifiers.lst').open(encoding='utf-8') as f:
+    with (get_cache_dir() / 'classifiers.lst').open(encoding='utf-8') as f:
         valid_classifiers = set(l.strip() for l in f)
-
-    invalid = classifiers - valid_classifiers
-    return ["Unrecognised classifier: {!r}".format(c)
-            for c in sorted(invalid)]
+    return valid_classifiers
 
 
-def _download_classifiers(cache_dir):
+def _download_and_chache_classifiers():
     """Get the list of valid trove classifiers from PyPI"""
     log.info('Fetching list of valid trove classifiers')
     resp = requests.get(
         'https://pypi.org/pypi?%3Aaction=list_classifiers')
     resp.raise_for_status()
 
-    with (cache_dir / 'classifiers.lst').open('wb') as f:
-        f.write(resp.content)
+    cache_dir = get_cache_dir()
+    try:
+        cache_dir.mkdir(parents=True)
+    except (FileExistsError, PermissionError, OSError):
+        # readonly mounted file raises OSError
+        pass
+
+    try:
+        with (cache_dir / 'classifiers.lst').open('wb') as f:
+            f.write(resp.content)
+    except (PermissionError, OSError):
+        # readonly mounted file raises OSError
+        # cache file could not be created
+        pass
+
+    valid_classifiers = set(l.strip() for l in resp.text.splitlines())
+    return valid_classifiers
+
+
+def _verify_classifiers(classifiers, valid_classifiers):
+    """Check classifiers against the downloaded list of known classifiers"""
+    invalid = classifiers - valid_classifiers
+    return ["Unrecognised classifier: {!r}".format(c)
+            for c in sorted(invalid)]
 
 
 def validate_classifiers(classifiers):
@@ -78,34 +87,34 @@ def validate_classifiers(classifiers):
 
     problems = []
     classifiers = set(classifiers)
-    with get_cache_dir() as cache_dir:
-        try:
-            problems = _verify_classifiers_cached(classifiers, cache_dir)
-        except FileNotFoundError as e1:
-            # We haven't yet got the classifiers cached
-            pass
-        else:
-            if not problems:
-                return []
-
-        # Either we don't have the list, or there were unexpected classifiers
-        # which might have been added since we last fetched it. Fetch and cache.
-
-        if os.environ.get('FLIT_NO_NETWORK', ''):
-            log.warning(
-                "Not checking classifiers, because FLIT_NO_NETWORK is set")
+    try:
+        valid_classifiers = _read_classifiers_cached()
+        problems = _verify_classifiers(classifiers, valid_classifiers)
+    except (FileNotFoundError, PermissionError) as e1:
+        # We haven't yet got the classifiers cached or couldn't read it
+        pass
+    else:
+        if not problems:
             return []
 
-        # Try to download up-to-date list of classifiers
-        try:
-            _download_classifiers(cache_dir)
-        except requests.ConnectionError:
-            # The error you get on a train, going through Oregon, without wifi
-            log.warning(
-                "Couldn't get list of valid classifiers to check against")
-            return problems
-        else:
-            return _verify_classifiers_cached(classifiers, cache_dir)
+    # Either we don't have the list, or there were unexpected classifiers
+    # which might have been added since we last fetched it. Fetch and cache.
+
+    if os.environ.get('FLIT_NO_NETWORK', ''):
+        log.warning(
+            "Not checking classifiers, because FLIT_NO_NETWORK is set")
+        return []
+
+    # Try to download up-to-date list of classifiers
+    try:
+        valid_classifiers = _download_and_chache_classifiers()
+    except requests.ConnectionError:
+        # The error you get on a train, going through Oregon, without wifi
+        log.warning(
+            "Couldn't get list of valid classifiers to check against")
+        return problems
+    else:
+        return _verify_classifiers(classifiers, valid_classifiers)
 
 
 def validate_entrypoints(entrypoints):
--- a/flit/validate.py
+++ b/flit/validate.py
@@ -63,7 +63,7 @@ def _download_and_cache_classifiers():
     try:
         with (cache_dir / 'classifiers.lst').open('wb') as f:
             f.write(resp.content)
-    except PermissionError:
+    except (PermissionError, FileNotFoundError):
         # cache file could not be created
         pass
     except OSError as e:
--- a/flit/validate.py
+++ b/flit/validate.py
@@ -1,5 +1,6 @@
 """Validate various pieces of packaging data"""
 
+import errno
 import io
 import logging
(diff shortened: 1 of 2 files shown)


tests/test_validate.py
```diff
@@ -107,3 +111,39 @@ def check(prurl):
...
+
+@pytest.fixture
+def patch_for_get_cache_dir():
+    # storing current environmental variables for resetting later
+    original_xdg = fv.os.environ.get("XDG_CACHE_HOME", None)
+    original_expanduser = fv.os.path.expanduser
+
+    # change the environmental variables for the test
+    fv.os.environ["XDG_CACHE_HOME"] = "/dev/null/nonexistent"
+    fv.os.path.expanduser = lambda x: "/dev/null/nonexistent"
```
takluyver: If you use the monkeypatch fixture, with its setattr & setenv methods, you don't need to do the cleanup manually. Which is good, because there's a minor bug - it won't clear XDG_CACHE_HOME if it was unset before.

https://docs.pytest.org/en/latest/monkeypatch.html

flit/validate.py
```diff
@@ -1,38 +1,49 @@
...
-        return Path(local, 'flit')
+
+    cache_dir = Path(local, "flit")
+    try:
+        cache_dir.mkdir(parents=True)
+    except FileExistsError:
+        pass
+    except (OSError, PermissionError):
+        cache_dir = Path(tempfile.TemporaryDirectory().name)
+        cache_dir.mkdir(parents=True)
```
takluyver: AFAIK, `TemporaryDirectory()` always creates the directory, so we don't need to do so here.

tests/test_validate.py
holgi: Thanks for pointing out the monkeypatch feature from pytest – I should read the docs more often. I was considering using pytest.mock but didn't want to introduce a new dependency. Nice to know, that pytest has also a simple patch feature built in. 

I changed the fixture to use the monkeypatch version in the new commit.

tests/test_validate.py
```diff
@@ -107,3 +111,39 @@ def check(prurl):
...
+    # reset the previously stored variables
+    if original_xdg is not None:
+        fv.os.environ["XDG_CACHE_HOME"] = original_xdg
+    fv.os.path.expanduser = original_expanduser
+
+
+# the following test will fail on windows, since windows allows the creation
+# of the directory "/dev/null/nonexistent/flit"
+
+@pytest.mark.skipif(os.name == 'nt', reason="Test will fail on windows")
```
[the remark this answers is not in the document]
holgi: Thanks for the hint, I changed this in a new commit.

flit/validate.py
holgi: I thought so too, but in one of my test setups this was not working reliably. Since I'm not at home and can't remember by heart which one it was,  I'd need to check on this later.

takluyver: Aha, I think I've got it. `TemporaryDirectory()` cleans itself up when the object is destroyed - which is immediately, because the code doesn't keep a reference to it. Use [mkdtemp](https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp) instead to get a directory which won't clear itself up.

Of course, this does mean each run of flit will create a new temp directory and not clear it up, which isn't ideal.

holgi: m(
Thanks for pointing this out. I ran head over heels into this kind of race condition in the first place and made things worse afterwards. Feeling kind of stupid now ;-).
Do you have any idea how to solve this without falling back on a module scope variable?

holgi: Unfortunately adding a new attribute to a pathlib.Path instance doesn't work.

takluyver: To do what you're currently doing, with a temporary directory that never gets cleaned up, the easiest thing is to use the `mkdtemp()` function instead of `TemporaryDirectory()`.

holgi: The other possibility would be - as you described - to use mkdtemp directly, but the cleanup afterwards must also be managed somehow.

holgi: Sorry for spamming you with comments. 
I think I have a fix in commit bef5f2b using a context manager. I had to slightly rewrite validate_classifiers also. #

holgi: @takluyver Sorry, I didn't see your reply – my connection is a little bit flaky, sitting on a train today. I actually want an automatic cleanup. Circumventing it was just a stupid mistake on my side.

takluyver:
That solves the cleanup problem, but if we create a new 'cache directory' each time the function is used, we're not really caching anything, so we can probably simplify things.

holgi:
I know that it's not caching anything. But this will only be in effect for (posix) users that don't have a (writable) home directory set, so probably related to some service / daemon.  And also only while using flit to install the project (not if the install is done via pip). 

Since I can't think of an posix path that has guaranteed write access for all users, I think not caching in this simple case is acceptable.

Currently an install in such a case fails, so this is an improvement IMHO.

Another possibility would be to introduce and check for a "FLIT_CACHE_DIR" environmental variable (similar to the suggestion in #320). In a case where there is no writable automagical cache path, the `flit` command could be prefixed like `FLIT_CAHCH_DIR=/vatr/tmp/cache flit install --pth-file`.

holgi:
Another possibility would be to move the check for `FLIT_NO_NETWORK` to the start of the function, so when `FLIT_NO_NETWORK` is set, the trove classifier check is passed completely.

takluyver:
Sorry to be unclear. I was thinking, if there's nothing being cached anyway, we don't need to bother creating a temporary directory and saving a file to disk - we can just get the data from the network and use it directly. If it's not clear how to do that, I can have a go myself next time I have a bit of spare time to work on it.

holgi:
Since I had some time this morning, I removed the caching. Should I add the changes to this pull request or if you want to close this I'll open a new one.

takluyver:
But it can still cache when there is a normal home directory, right?

If you're happy with the changes, feel free to push them to this PR. Or if you're not sure, it's fine to make another PR and we can work out what's best.

holgi:
With the changes I made today (not pushed so far), caches would be created and used as before if it is possible.

After a download of the classifiers an attempt is made to create a cache file for later use, but it doesn't matter if this is successful or not.

flit/validate.py
```diff
@@ -34,17 +34,15 @@ def get_cache_dir() -> Path:
                 or os.path.expanduser('~\\AppData\\Local')
         return Path(local, 'flit')
 
-def _verify_classifiers_cached(classifiers):
+
+def _read_classifiers_cached():
     """Check classifiers against the downloaded list of known classifiers"""
```
takluyver: The docstring here needs to be updated.

flit/validate.py
```diff
@@ -34,17 +34,15 @@ def get_cache_dir() -> Path:
...
         valid_classifiers = set(l.strip() for l in f)
-
-    invalid = classifiers - valid_classifiers
-    return ["Unrecognised classifier: {!r}".format(c)
-            for c in sorted(invalid)]
+    return valid_classifiers
 
 
-def _download_classifiers():
+def _download_and_chache_classifiers():
```
takluyver:
```suggestion
def _download_and_cache_classifiers():
```

flit/validate.py
```diff
@@ -54,10 +52,27 @@ def _download_classifiers():
     cache_dir = get_cache_dir()
     try:
         cache_dir.mkdir(parents=True)
-    except FileExistsError:
+    except (FileExistsError, PermissionError, OSError):
+        # readonly mounted file raises OSError
```
takluyver: I don't really like catching any OSError - that's quite a broad category. OSError objects normally have an `e.errno` attribute which you can check against constants in the [`errno` module](https://docs.python.org/3/library/errno.html) to look for specific errors. E.g. it sounds like this might be `EROFS` (read only file system) - but check if that's what you're hitting.

This pattern was more common in Python 2 code, before subclasses like `FileExistsError` and `PermissionError` were split out as their own subclasses.

tests/test_validate.py
```diff
@@ -107,3 +108,74 @@ def check(prurl):
...
+            def __truediv__(self, other):
+                return self
+            def __enter__(self):
+                return ["A", "B", "C"]
+            def __exit__(self, *args):
+                pass
+
+        return MockChache()
+
+    monkeypatch.setattr(fv, "get_cache_dir", mock_get_cache_dir)
```
takluyver: I don't much like this kind of complex mocking. It ties the test closely to how the function is implemented - e.g. it assumes the code does `with path.open() as f:` and iterates over f, rather than something like `path.read_text()`.

For tests that deal with file access, I prefer to make a temporary directory and mock things to point to that. Pytest has a handy `tmp_path` fixture which creates a temp dir and gives the test a pathlib object for it.

tests/test_validate.py
```diff
@@ -107,3 +108,74 @@ def check(prurl):
...
+
+    classifiers = fv._read_classifiers_cached()
+
+    assert classifiers == {'A', 'B', 'C'}
+
+
+def test_download_and_chache_classifiers():
+    classifiers = fv._download_and_chache_classifiers()
```
takluyver: In general, I like it to be possible to run the tests without requiring network access. There are two approaches for this:

1. Mark tests which need network access, so there's a clear way to exclude them. I don't think there are any examples in Flit, but here's one from another of my projects: https://github.com/takluyver/pynsist/blob/bf2408bbd01b5da63b93107f48435a5a57f9048b/nsist/tests/test_pypi.py#L11-L14
2. Mock out the network responses, e.g. see the `test_upload` module.

holgi:
Thanks for your patience with me and my changes ;-)

flit/validate.py
holgi: This is fixed in commit 909f7b4

flit/validate.py
holgi: Typo is fixed in commit 909f7b4

flit/validate.py
holgi: I was running into this problem on a Mac with 10.15 "Catalina" running Python 3.7. The root and other system directories are read-only. 

Trying to create the dir "/nonexistent/" raises a `OSError` for a readonly file system – unfortunately not a subclassed error like `PermissionError` but an `OSError` itself:

```
>>> open("/foo", "w")
  File "<stdin>", line 1, in <module>
OSError: [Errno 30] Read-only file system: '/foo'
```

I changed the try-except-block to be a little bit more verbose and check for an explicit `errno.EROFS`

tests/test_validate.py
holgi: For `test_read_classifiers_cached()` and `test_download_and_cache_classifiers()` I use the suggested pytest `tmp_path` fixtures.

For the `test_download_and_cache_classifiers_with_unacessible_dir()` test I couldn't come up with a good idea without a custom mock class, but I think I stripped it down to a bare minimum.

tests/test_validate.py
holgi: Since you used `responses` already, I've choosen the second option.

tests/test_validate.py
```diff
@@ -107,3 +111,111 @@ def check(prurl):
...
+
+@responses.activate
+@pytest.mark.parametrize("error", [PermissionError, OSError(errno.EROFS, "")])
+def test_download_and_cache_classifiers_with_unacessible_dir(monkeypatch, error):
+    responses.add(
+        responses.GET,
+        'https://pypi.org/pypi?%3Aaction=list_classifiers',
+        body="A\nB\nC")
+
+    class MockChacheDir:
```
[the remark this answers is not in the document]
holgi: Fixed in the next commit

tests/test_validate.py
```diff
@@ -107,3 +111,111 @@ def check(prurl):
...
+        def mkdir(self, parents):
+            raise error
+        def __truediv__(self, other):
+            raise error
+
+    monkeypatch.setattr(fv, "get_cache_dir", MockChacheDir)
```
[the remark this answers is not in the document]
holgi: Fixed in the next commit

tests/test_validate.py
```diff
@@ -107,3 +111,111 @@ def check(prurl):
...
+def test_download_and_cache_classifiers_not_catching_oserror_mkdir(monkeypatch):
+    responses.add(
+        responses.GET,
+        'https://pypi.org/pypi?%3Aaction=list_classifiers',
+        body="A\nB\nC")
+
+    unused_error_nr = max(errno.errorcode) + 1
+    class MockChacheDir:
+        def mkdir(self, parents):
+            raise OSError(unused_error_nr, "")
```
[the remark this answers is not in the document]
holgi: Well, I overdid it just to have the coverage report pass ¯\_(ツ)_/¯
I can remove it if it annoys you ;-)

holgi:
Sure no problem. You are probably doing this in your spare time 😄 👍

tests/test_validate.py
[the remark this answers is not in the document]
takluyver: Thanks, I think I'd prefer to drop these two tests. Coverage is a useful metric, but I don't aim for 100% coverage, and I don't mind if a PR decreases coverage a bit if the alternative is writing really contorted tests.

takluyver:
I am indeed doing this in my spare time - though there might be more spare time with all these social distancing measures.

holgi:
I removed the two tests as you proposed. Thanks again for your work and the patience with this pull request :-)

takluyver:
Thanks for your patience with my reviewing. :slightly_smiling_face:

holgi:
👍

PR #320, opened 2020-02-23
use HOME to resolve the cache directory

LnL7:
This enables overriding the location, for example if flit runs as a
build user without a home directory like /var/empty.

### DIFF
--- a/flit/validate.py
+++ b/flit/validate.py
@@ -18,15 +18,14 @@ def get_cache_dir() -> Path:
     Does not ensure that the cache directory exists.
     """
     # Linux, Unix, AIX, etc.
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks, but I think `expanduser()` already respects HOME if it's set:

https://github.com/python/cpython/blob/c3fa634096eedbaf477698adab666f03085a7928/Lib/posixpath.py#L252

There's actually another PR open at the moment to handle the same case by falling back to a temporary directory (#319).

LnL7:
Oh interesting, I assumed it was just a frontend for `getpwnam`. And not depending on a cache directory at all would be even better.

PR #327, opened 2020-03-20
Update example pyproject.toml, fix requires-python (fixes #326)

### LINKED ISSUE #326: Example pyproject.toml is invalid
https://github.com/takluyver/flit/blob/f8afe15b183c03c3dd091ebb4825af7944753b5b/doc/pyproject_toml.rst#L111-L134

Using the provided example, running `flit install` produces the error:
```
Invalid requires-python: '3'.                E-flit.validate
Invalid config values (see log)
```

The example is now quite outdated from Flit's `pyproject.toml`. We can resync it, although it may get out of date again in the future. The actual error can be resolved by updating the `requires-python` to `>= 3`.

Edward-Knight:
Closes #326

### DIFF
--- a/doc/pyproject_toml.rst
+++ b/doc/pyproject_toml.rst
@@ -118,12 +118,13 @@ Here's the full metadata section from flit itself:
     author-email="thomas@kluyver.me.uk"
     home-page="https://github.com/takluyver/flit"
     requires=[
+        "flit_core>=2.2.0",
         "requests",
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks for catching this. Longer term, I may use another project's `pyproject.toml` as an example to simplify things, but it's good to have it actually valid in the meantime. :-)

PR #333, opened 2020-04-07
Generated setup.py use setuptools when available

kitterma:
It would be nice if the generated setup.py would support using setuptools.  The generated setup.py already includes setuptools specific options, so I think it would make sense for it to use setuptools when it available (example from pep517):

> Unknown distribution option: 'install_requires'
> Unknown distribution option: 'extras_require'

This PR changes the generated setup.py to use setuptools when available, but fall back to distutils (current behavior) if it is not.


takluyver:
Thanks, but this is deliberate - the `setup.py` is meant as a temporary shim so older versions of pip can handle the sdist. I want to discourage running it directly. Pip ensures that setuptools is loaded even if the script doesn't explicitly specify it, and works around some of the problems that setuptools otherwise causes.

I'll add a comment to this effect.

takluyver:
Added a comment in 8ed3317f0121ab21ff3f5f8ddacb8bb64952188a

Issue #313, opened 2020-01-31
Compute SOURCE_DATE_EPOCH automatically (e.g., from git)


paulproteus:
Hi @takluyver !

I was reading https://flit.readthedocs.io/en/latest/reproducible.html , and I was excited to see that reproducible builds are something you've considered with `flit`. I wondered if there was a way to make reproducibility automatic.

Toward that goal, I'm interested in working with you to design a user flow where `flit` can compute the `SOURCE_DATE_EPOCH` from git. I'm also interested in submitting a patch that implements this, with an approximate ETA of 2-3 weeks from when we decide how it should work, if that's helpful. I also have some other questions about this, which you'll find below.

I can think of a few implementation strategies:

- `flit` could automatically check the `git log` for the most recent commit, and use the timestamp as the `SOURCE_DATE_EPOCH`, if this is a git repository (using the same logic as `flit` already uses).

- `flit publish` (and/or `flit build`) could accept a flag called e.g. `--source-date-from-vcs`, so that users have to opt-in to this behavior.

- `flit publish` (and/or `flit build`) could accept a flag called `--use-timestamps`, in which case, we use the current behavior of embedding timestamps (subject to `SOURCE_DATE_EPOCH`). If that flag is omitted, we use some zero timestamp as the timestamp in wheels and tar files.

I have a few related questions:

- The doc at https://flit.readthedocs.io/en/latest/reproducible.html seems to imply that a `sdist` doesn't depend on the current time. My experiments show that the `sdist` built by `flit build` do contain timestamps of files. Should I change that doc? Happy to do so. (Context: I did `flit build` on a clone of this repo, then did `tar -t -v --full-time -f dist/flit-2.2.0.tar.gz | grep 2020-01-31 | head -n1` to confirm that I got a value from today, even though the repo is unchanged since a couple of weeks ago.)

- `flit` is embedding a timestamp of 2016-01-01 for its generated files in the wheel <https://github.com/takluyver/flit/blame/93482e44d804b80c770203ec6c2e375170ece9ef/flit_core/flit_core/wheel.py#L173> but not for the user's code. Would it be better and simpler to use this 2016-01-01 timestamp for user code, too? This would also address my goal here, I believe. :)

- If we go ahead with the `git`-based implementation strategy, would you want this to work equally well for `hg` as part of merging the patch? If so, I can commit to adding that to my patch before submitting it for your final review.

Happy to chat more, either here or somewhere lower-latency if you desire.

[...]

PR #335, opened 2020-04-12
PEP 610 (direct_url.json) support

sbidoul:
Initial PEP 610 support is merged in pip master (will be in pip 20.1). This PR is meant to add PEP 610 support to flit.

- [x] support for `flit install --symlink/--pth-file`
- [x] suport for `flit install`
- [x] tests

Currently `flit install` (without --symlink/--pth-file) builds a wheel then invokes `pip install {temporary wheel file}`. This means pip will consider the wheel is being installed from a direct url which is actually a temporary file and will record that temporary url. The desired behaviour is to record the url of the local project directory.

To achieve that desired behaviour, there are two options:

1. Change `install_with_pip()` to merely do `pip install {self.directory}` and let pip do the PEP 517 build and install. That is the simplest approach, but has the drawback of possibly invoking another version of `flit` as build backend, and being somewhat slower. 
2. Keep `install_with_pip()` as is and patch the generated `direct_url.json` to point to the local project directory. This is somewhat complex because the `direct_url.json` that has been generated by pip has to be found correctly, and its hash in the `RECORD` file has to be updated.

### DIFF
--- a/flit/install.py
+++ b/flit/install.py
@@ -4,6 +4,7 @@ import logging
 import os
 import os.path as osp
 import csv
+import json
 import pathlib
 import random
 import shutil
@@ -384,6 +385,16 @@ class Installer(object):
             with (dist_info / 'entry_points.txt').open('w') as f:
                 common.write_entry_points(self.ini_info.entrypoints, f)
             self.installed_files.append(dist_info / 'entry_points.txt')
+
+        with (dist_info / 'direct_url.json').open('w', encoding='utf-8') as f:
+            json.dump(
+                {
+                    "url": self.directory.as_uri(),
+                    "dir_info": {"editable": bool(self.symlink or self.pth)}
+                },
+                f
+            )
+        self.installed_files.append(dist_info / 'direct_url.json')
 
         with (dist_info / 'RECORD').open('w', encoding='utf-8') as f:
             cf = csv.writer(f)
--- a/flit/install.py
+++ b/flit/install.py
@@ -329,34 +329,26 @@ class Installer(object):
         self.write_dist_info(dirs['purelib'])
 
     def install_with_pip(self):
+        """Let pip install the project directory
+
+        pip will create an isolated build environment and install build
+        dependencies, which means downloading flit_core from PyPI. We ask pip
+        to install the project directory (instead of building a temporary wheel
+        and asking pip to install it), so pip will record the project directory
+        in direct_url.json.
+        """
         self.install_reqs_my_python_if_needed()
-
-        with tempfile.TemporaryDirectory() as td:
-            temp_whl = osp.join(td, 'temp.whl')
-            with open(temp_whl, 'w+b') as fp:
-                wb = WheelBuilder(
-                    str(self.directory),
-                    self.module,
-                    metadata=common.make_metadata(self.module, self.ini_info),
-                    entrypoints=self.ini_info.entrypoints,
-                    target_fp=fp,
-                )
-                wb.build()
-
-            renamed_whl = osp.join(td, wb.wheel_filename)
-            os.rename(temp_whl, renamed_whl)
-            extras = self._extras_to_install()
-            extras.discard('.none')
-            whl_with_extras = '{}[{}]'.format(renamed_whl, ','.join(extras)) \
-                if extras else renamed_whl
-
-            cmd = [self.python, '-m', 'pip', 'install', whl_with_extras]
-            if self.user:
-                cmd.append('--user')
-            if self.deps == 'none':
-                cmd.append('--no-deps')
-            shell = (os.name == 'nt')
-            check_call(cmd, shell=shell)
+        extras = self._extras_to_install()
+        extras.discard('.none')
+        req_with_extras = '{}[{}]'.format(self.directory, ','.join(extras)) \
+            if extras else str(self.directory)
+        cmd = [self.python, '-m', 'pip', 'install', req_with_extras]
+        if self.user:
+            cmd.append('--user')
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks, and congratulations on getting a PEP to provisional acceptance.

I think I favour approach 1 (letting pip handle the installation from the source tree), but as flit is already geared up to detect & interact with git & hg, I wonder if it makes sense to tell pip about it as a VCS repository (if the directory is one, of course). When you're doing a normal (non-editable) install, it seems better to record the commit that got installed. But then if there are uncommited changes... :confused: Thoughts?

If we do go that way, I'd still ensure that flit loads and validates the configuration before invoking pip - when `flit_core` is called as a backend to build the wheel, it will skip most of the validation, on the grounds that that's useful when you're developing the package, not when someone tries to install it.

sbidoul:
Thanks! 

`pip install .` will record the local directory, so flit doing the same would be natural.

Once the local directory is recorded, a command such as pip freeze can try to do smart things such as detecting if the local directory and possibly convert it to a VCS reference. pip does it for editable and it's not exactly trivial, for instance when there are multiple origins. Also there was a recent feature request asking for pip freeze to *not* try to be smart with VCS directories: https://github.com/pypa/pip/issues/7896

So yes, I think local directory is less surprising (and simpler to implement), while leaving the door open to more elaborate use cases.

takluyver:
The obvious issue with `pip freeze` finding the commit is that the repository may have been updated since you installed it. Converting it to a VCS reference at install time means you can capture the commit that gets installed. (I'm only talking about normal installs here - editable installs are probably a mess for pip freeze whatever you do).

I wouldn't deal with git remotes, just turn the directory path into a git URL `git+file:///...` so that pip uses git to install from there.

The one bit I'm not sure about is what it should do if there are uncommitted changes. Maybe that's enough to invalidate the whole idea, but it's still kind of appealing to automatically capture the commit when that's possible.

takluyver:
Or maybe we should focus on reproducible builds (e.g. see discussion on #313), and continue generating a wheel, so that pip records the hash of something we can recreate. But that's still not much use without a way to recreate/retrieve a wheel from a hash.

takluyver:
I haven't come up with a neat way to deal with uncommitted changes, and on reflection, I think it makes sense for `pip install .` and `flit install` to behave similarly. So let's go ahead without trying to be smart about capturing the VCS metadata.

I still think this is worth giving more thought - when I install from a local directory, it's often a git repo with no uncommitted changes. It seems like a wasted opportunity not to record the commit ID - especially as pip already knows about git repositories, and the PEP defines a format for that information. But I won't try to work around that in Flit, at least for now. :wink:

sbidoul:
@takluyver this is now ready for review.

> I'd still ensure that flit loads and validates the configuration before invoking pip

This is the case, via `read_flit_config`.

BTW, during testing I noticed that [some pyproject.toml](https://github.com/takluyver/flit/blob/c18420eaa48f5c2a4b4f0411a25cc0a62c397335/tests/samples/extras/pyproject.toml#L1-L2) in the test suite have no `build-backend` defined. pip complains about it, while `flit`'s validation routine is happy with it. I'm not sure if more validation is required there.

tests/test_install.py
```diff
@@ -27,10 +28,28 @@ def setUp(self):
...
+            / '{}-{}.dist-info'.format(package, version)
+            / 'direct_url.json'
+        )
+        assert_isfile(direct_url_file)
+        with direct_url_file.open() as f:
+            direct_url = json.load(f)
+            assert direct_url.get('url').startswith('file:///')
+            assert direct_url.get('url') == directory.as_uri()
+            assert 'dir_info' in direct_url
+            assert direct_url.get('dir_info').get('editable') is expected_editable
```
takluyver, line 44:
```suggestion
            assert direct_ur['url'].startswith('file:///')
            assert direct_url['url'] == directory.as_uri()
            assert 'dir_info' in direct_url
            assert direct_url['dir_info']['editable'] is expected_editable
```

IMO, if the key is ever missing, a KeyError makes that clear better than getting `None` and trying to check it.

flit/install.py
```diff
@@ -329,33 +330,17 @@ def install_directly(self):
...
-                cmd.append('--user')
-            if self.deps == 'none':
-                cmd.append('--no-deps')
-            shell = (os.name == 'nt')
-            check_call(cmd, shell=shell)
+        extras = self._extras_to_install()
+        extras.discard('.none')
+        req_with_extras = '{}[{}]'.format(self.directory, ','.join(extras)) \
+            if extras else str(self.directory)
+        cmd = [self.python, '-m', 'pip', 'install', req_with_extras]
```
takluyver, line 337: Let's have a comment to explain about using pip to do the installation, and that it will install `flit_core` in a temporary environment and invoke it there to make a wheel.

takluyver:
Thanks, this LGTM, I've just got a couple of minor suggestions.

tests/test_install.py
sbidoul: done

flit/install.py
sbidoul: I added some explanations in the docstring.

takluyver:
Thanks!

takluyver:
Trying this interactively turned up a problem that the tests weren't catching, when `self.directory` is not absolute. #341 should fix that.

sbidoul:
Oh, yes good catch. Thanks!

Now I'll see if I can add direct URL support to `pip list`, because what it currently shows in the `Location` column is where the metadata is located, which is not super useful to locate editable installs.

PR #337, opened 2020-04-18
When doing `flit init`, if readme file is available, add it to pyproject.toml

sander76:
When doing `flit init`, flit will look for a readme file in the root of the project. (`readme.md`, `readme.rst`, `readme.txt`, lower or uppercase). If found it will ask the user to include it in the `pyproject.toml` file as an `description-file` entry.

### DIFF
--- a/flit/init.py
+++ b/flit/init.py
@@ -126,6 +126,13 @@ class IniterBase:
         with (self.directory / 'LICENSE').open('w', encoding='utf-8') as f:
             f.write(license_text.format(year=year, author=author))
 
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks, this makes sense. I'd actually skip the prompt and automatically use the file if one is found. It's easy to delete the line from `pyproject.toml` afterwards if it's not what you want.

sander76:
@takluyver Ok. I'll make the change to just include it automatically.

takluyver:
Thanks!

PR #355, opened 2020-06-14
test: add test for building with unicode filenames

bpabel:
Added an explicit test for building against unicode filenames that I neglected to add with the last PR.


tests/test_build.py
```diff
@@ -69,3 +69,25 @@ def test_build_module_no_docstring():
...
+def test_build_package_with_unicode(copy_sample):
+    list_files_template = """\
+#!{python}
+import sys
+from os.path import join
+if '--deleted' not in sys.argv:
+    files = [
+        'pyproject.toml', 
+        'packageunicode/__init__.py', 
+        u'packageunicode/No\\u00ebl.jpg',
```
takluyver, line 82:
```suggestion
        'packageunicode/No\\u00ebl.jpg',
```

We don't need to worry about Python 2 here. :-)
We're only

bpabel: I’m trying to figure out why the appveyor ci is failing.  The tests all run successfully locally with tox.

mohdzahidabdullah:
> Added an explicit test for building against unicode filenames that I neglected to add with the last PR.

https://github.com/pypa/flit/pull/355#issue-638291555

takluyver:
Let's see how this goes with the current test pipeline.

takluyver:
Still no dice:

```
   E               FileNotFoundError: [WinError 2] The system cannot find the file specified: 'C:\\Users\\runneradmin\\AppData\\Local\\Temp\\pytest-of-unknown\\pytest-0\\test_build_package_with_unicod0\\packageunicode\\packageunicode\\NoÃ«l.jpg'
```

Issue #353, opened 2020-06-12
Ability to write Requires-External metadata


rmorshea:
This may be out of scope for the project, but I'd like to be able to write to the `Requires-External` field of the package `METADATA`. If this isn't a feature that can be provided out of the box, I was already planning to wrap `flit_core`, so if there's at least some sanctioned way to overwrite the internal behavior of [`Metadata.write_metadata_file`](https://github.com/takluyver/flit/blob/master/flit_core/flit_core/common.py#L303) that would be great.

takluyver:
My initial inclination is to say it's out of scope - I'm not aware of any tools that consume it, and for telling humans about external requirements, I think you're better off describing them in free text in the description. But feel free to make a case for it - maybe there's some reason to use it that I'm not aware of.

Be aware that the only public interfaces to Flit are the PEP 517 API (`flit_core.buildapi`) and the command line interface. Everything else is implementation, and can be changed with no deprecation period or warning in the release notes, even in minor versions. Maintaining an extensible API is a bigger challenge.

bpabel:
Since `flit_core` is responsible for writing the package metadata, wouldn't it be in scope to support all of the [core metadata](https://packaging.python.org/specifications/core-metadata/) fields?

rmorshea:
@takluyver I was planning on using it to specify Javascript dependencies that need to be installed (e.g. for a Jupyter widget) that would be consumed by a downstream installer. So instead of having to run:

```
pip install my-widget
jupyter labextension install client-for-my-widget
```

You could simply run:

```
jupyter labextension install my-widget
```

Where `jupyter labextension install` is aware of the `Requires-External` metadata so it can `pip install` the package and then automatically install the client.

Admittedly though this is a niche use case, and as much as I wish this could be added, it looks like `setuptools` opted not to include it in their API either: https://github.com/pypa/setuptools/pull/631#issuecomment-231545348

rmorshea:
If this field isn't natively supported by flit I can see two options that would allow me to move forward:

1. `flit_core` could define a generic API that would allow my (or others) to manipulate the package files before build. Something like:

```python
with flit_core.wheel_build_context() as (builder, build_dir):
    builder.prepare()
    with (build_dir / "METADATA").open("a") as meta_file:
        # my custom prep code
        ...
    # make wheel from prepared files
    builder.build()
```

2. I can vendor my own version of flit.

rmorshea:
[...]

PR #356, opened 2020-06-14
feat: add support for Requires-External metadata

bpabel:
Support for #353


rmorshea:
@takluyver any new thoughts on this?

wimglenn:
@takluyver  I would also like to see this feature

PR #359, opened 2020-06-27
Update generated scripts to match pip installed scripts

ksunden:
I have some console scripts which are class methods (the main remains the same, but I have a hierarchy of subclasses in different packages which fill out functionality for several different entry points)

```
[tool.flit.scripts]
foo="bar.baz:Spam.main"
```

This caused flit to fail when doing editable installs (symlink/pth-file), but work when doing "regular" installs (via pip)

I found that the template used here expected _only_ a single part function call.

This PR takes the [template as used by pip (provided by the vendored distlib)](https://github.com/pypa/pip/blob/master/src/pip/_vendor/distlib/scripts.py#L41) and modifies the functions that interact with the template to provide the imported name separate from the function name.

### DIFF
--- a/flit_core/flit_core/common.py
+++ b/flit_core/flit_core/common.py
@@ -209,27 +209,32 @@ def check_version(version):
 
 script_template = """\
 #!{interpreter}
-from {module} import {func}
+# -*- coding: utf-8 -*-
+import re
+import sys
+from {module} import {import_name}
 if __name__ == '__main__':
-    {func}()
+    sys.argv[0] = re.sub(r'(-script\.pyw|\.exe)?$', '', sys.argv[0])
+    sys.exit({func}())
 """
 
 def parse_entry_point(ep):
     """Check and parse a 'package.module:func' style entry point specification.
 
-    Returns (modulename, funcname)
+    Returns (modulename, import_name, funcname)
     """
     if ':' not in ep:
         raise ValueError("Invalid entry point (no ':'): %r" % ep)
     mod, func = ep.split(':')
 
-    if not func.isidentifier():
-        raise ValueError("Invalid entry point: %r is not an identifier" % func)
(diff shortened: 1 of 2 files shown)


ksunden:
Looks like red CI is because I'm missing one line in codecov, should I write a new test that covers that one line?

takluyver:
Thanks, and sorry I didn't look at this before.

I'm not worried about the coverage; it's useful to know what tests are exercising overall, but I don't aim for 100% coverage.

I think it would be slightly neater to leave `parse_entry_point` returning the same 2-tuple it already did, and make `import_name` in `install_scripts()`. But other than that, this looks good.

ksunden:
Updated!

I had debated doing it that way in the first place.

flit_core/flit_core/common.py
```diff
@@ -209,22 +209,27 @@ def check_version(version):
...
-    {func}()
+    sys.argv[0] = re.sub(r'(-script\\.pyw|\\.exe)?$', '', sys.argv[0])
+    sys.exit({func}())
 """
 
 def parse_entry_point(ep):
     """Check and parse a 'package.module:func' style entry point specification.
 
-    Returns (modulename, funcname)
+    Returns (modulename, import_name, funcname)
```
takluyver, line 224:
```suggestion
    Returns (modulename, funcname)
```

takluyver:
Thanks @ksunden !

ksunden:
Any chance I could poke you for a release which includes this? I have some people getting started developing on the system I ran into this on, and would prefer if I could just point them to released flit, rather than having to either avoid editable installs or point to the git version.

takluyver:
Just released Flit 3.0. Let's see what breaks. :wink:

ksunden:
Thanks!

PR #360, opened 2020-07-07
add spaces to warning about missing __version__

bast:
Because a likely reaction to this useful warning is to copy paste this into the module and adjust x, y, and z but then it's convenient to not have to add spaces in addition.


takluyver:
Thanks.

PR #366, opened 2020-08-16
Deterministically produce RECORD file

vexelnet:
Hi,

While working on a Debian package that uses flit, I noticed that the package could not be built reproducibly (https://reproducible-builds.org).

This patch will aid to reproducibly build Debian packages of projects that use flit.

### DIFF
--- a/flit/install.py
+++ b/flit/install.py
@@ -392,7 +392,7 @@ class Installer(object):
 
         with (dist_info / 'RECORD').open('w', encoding='utf-8') as f:
             cf = csv.writer(f)
-            for path in self.installed_files:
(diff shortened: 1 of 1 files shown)


flit/install.py
```diff
@@ -392,7 +392,7 @@ def write_dist_info(self, site_pkgs):
 
         with (dist_info / 'RECORD').open('w', encoding='utf-8') as f:
             cf = csv.writer(f)
-            for path in self.installed_files:
+            for path in sorted(self.installed_files, key=lambda s: str(s)):
```
takluyver, line 395:
```suggestion
            for path in sorted(self.installed_files, key=str):
```

takluyver:
Thanks

PR #368, opened 2020-08-18
Fix newline for RECORD

### LINKED ISSUE #367: flit install --pth-file breaks conda list
I tried to use `flit install --pth-file` on Windows and encountered the same problem as described [here](https://github.com/conda/conda/issues/9074).

flying-sheep:
Fixes #367, fixes conda/conda#9074

See this footnote: https://docs.python.org/3/library/csv.html#id3

### DIFF
--- a/flit/install.py
+++ b/flit/install.py
@@ -390,7 +390,7 @@ class Installer(object):
             )
         self.installed_files.append(dist_info / 'direct_url.json')
 
-        with (dist_info / 'RECORD').open('w', encoding='utf-8') as f:
+        with (dist_info / 'RECORD').open('w', encoding='utf-8', newline='') as f:
             cf = csv.writer(f)
             for path in self.installed_files:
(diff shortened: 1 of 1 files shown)


flit/install.py
```diff
@@ -390,7 +390,7 @@ def write_dist_info(self, site_pkgs):
             )
         self.installed_files.append(dist_info / 'direct_url.json')
 
-        with (dist_info / 'RECORD').open('w', encoding='utf-8') as f:
+        with (dist_info / 'RECORD').open('w', encoding='utf-8', newline='') as f:
```
takluyver, line 393:
```suggestion
        # newline='' because the csv module does its own newline translation
        with (dist_info / 'RECORD').open('w', encoding='utf-8', newline='') as f:
```

takluyver:
Thanks Philipp

PR #376, opened 2020-11-11
Switch Linux CI from Travis to Github Actions

takluyver:
Travis is [going to be less generous with open source time](https://blog.travis-ci.com/2020-11-02-travis-ci-new-billing), and it's already limiting capacity enough to create a substantial backlog. Github Actions seems pretty popular, so let's try that.

We're using Tox, so it's not too much work to switch from one CI provider to another.


tannewt:
I've been using Github Actions a while and see no issues. Good work!

takluyver:
Thanks!

PR #381, opened 2020-11-30
Added support for --extra-index-url to flit install

alesanfra:
This PR adds support for the option `--extra-index-url` to `flit install`. This will make possible to use flit also with private indexes (widely used in business environment) or test repositories (e.g. test.pypi.org). The implemented behavior is same of `pip install  --extra-index-url <url>`.

Example:
`flit install --extra-index-url https://test.pypi.org/simple`

### DIFF
--- a/flit/__init__.py
+++ b/flit/__init__.py
@@ -124,6 +124,9 @@ def main(argv=None):
         help="Install the dependencies of these (comma separated) extras additionally to the ones implied by --deps. "
              "--extras=all can be useful in combination with --deps=production, --deps=none precludes using --extras"
     )
+    parser_install.add_argument('--extra-index-url', default=None, type=str,
+        help="Extra URL of package indexes to use in addition to pypi.org"
+    )
 
     # flit init --------------------------------------------
     parser_init = subparsers.add_parser('init',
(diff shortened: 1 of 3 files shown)


sander76:
I found another option to do this: use the extra-index-url in an environment variable, or pip config file.

takluyver:
Thanks for contributing, and sorry it's taken me so long to look at this!

I'm inclined not to add options to `flit install`, for a couple of reasons. Firstly, there's less need for `flit install` to exist at all now that `pip install .` works for packages using Flit - it's basically only there for the editable install options, which don't yet have a standardised hook so pip can request them. So it might be that eventually, `flit install` goes away altogether.

Secondly, `flit install` is actually using pip, and all pip's options can be [controlled by environment variables](https://pip.pypa.io/en/stable/user_guide/#environment-variables) - in this case, `PIP_EXTRA_INDEX_URL`. So I'd rather not add a whole load of options (see `pip install --help`) which just get passed through to the pip commands, when everything they'd do is possible anyway.

My feeling is that extra-index-url isn't special enough to make an exception for it. If you want to make a case that it *is* special, fire away - but I'm starting out sceptical. :slightly_smiling_face:

alesanfra:
Your reasoning makes perfect sense to me, either `pip install . --extra-index-url internal.repo` or environment variables are the way to go at this point.
Thanks for the detailed answer, I think we can discard this pull request.

takluyver:
Thanks @alesanfra !

Issue #141, opened 2017-10-19
Runtime requirements may be necessary at build time


zenbot:
Apologies first off if this has been discussed before — I've only had a quick scan through old issues to see if this has already been mentioned.

`flit`'s behaviour of importing the package being built to get its description and version means that any runtime requirements in the top level of the package must be available for import in the build environment. This can be painful when the runtime requirements aren't trivially installable.

My first thought on this was, "these should be just made optionally configurable in flit.ini", but this goes against flit's (excellent!) there-should-be-one-way-to-do-it ethos.

My second thought was "we could just use `ast` to pull the docstring and version out", but that assumes that `__version__` is defined in the `__init__.py` or module, and not imported from elsewhere.

Do you have any thoughts on this? Would you consider allowing optional configuration of these things in `flit.ini`, or `ast`-walking as a fallback if the package fails to be imported? Or do you consider the package being importable at build time a hard requirement?

zenbot:
One advantage of allowing optional configuration of the description and version in `flit.ini` that I just thought of: it enables the building of PEP 420-style native namespace packages.

Carreau:
Thanks for open this on your though on how to improve flit and the kind words !

This seem similar to one of [the earlier issue](https://github.com/takluyver/flit/pull/63) I opened. 

I think there are many point to take into consideration when deciding what features to have in flit, and it's always a right balance, and as you pointed out one of flit quality is the strong opinions that @takluyver baked into it. 

One things I am curious about is in which case will you built a package when you cannot import it ? One thing I can think of is in an automated building process which is only responsible for building. 
Now one of the question I have is should you build a package if you can't test it or at least import it before building ? It feel to me like in most case the release process should be : 
  - test the pacakge
  - if test pass: package. 

Now there is already a use-case where this breakdown : flit can build a wheel for Python2, but that does not mean you can test on Python 2. 

The other two side are the complexity of implementation, it's cost; vs the number of case where it is actually useful. 

In this case, without amore explicit example, I have the feeling that the cost-benefit ratio would not be that great for the use case; in particular because I _feel_ like the env that build the wheel should be the one testing it. 

On the other hand I'm not sure there are many packages that actually have dynamic generation of version number; so I'm unsure how many packages would be broken by such an implementation. 

[...]

PR #382, opened 2020-12-03
Allow getting version despite deps missing

flying-sheep:
Also gets rid of deprecated methods

### DIFF
--- a/flit_core/flit_core/common.py
+++ b/flit_core/flit_core/common.py
@@ -150,13 +150,19 @@ def get_docstring_and_version_via_import(target):
     from it.
     """
     log.debug("Loading module %s", target.file)
-    from importlib.machinery import SourceFileLoader
-    sl = SourceFileLoader(target.name, str(target.file))
+    from importlib.machinery import PathFinder
+    from importlib.util import spec_from_file_location, module_from_spec
+    spec = spec_from_file_location(target.name, str(target.file))
+    module = module_from_spec(spec)
+    error = None
     with _module_load_ctx():
-        m = sl.load_module()
+        try:
+            spec.loader.exec_module(module)
+        except ImportError as e:
+            error = e
     docstring = m.__dict__.get('__doc__', None)
     version = m.__dict__.get('__version__', None)
-    return docstring, version
+    return docstring, version, error
 
 
 def get_info_from_module(target):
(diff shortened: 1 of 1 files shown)


flying-sheep:
@takluyver opinions?

flying-sheep:
@takluyver what do you think?

takluyver:
Hi @flying-sheep, sorry it's taken me ages to look at these PRs (this and #374). I know I'm rather neglecting my role as a maintainer of Flit - when I get a bit of time to do non-work programming, there's generally something more either more fun or where I feel I can make more concrete progress in a small chunk of time.

I like this idea better than #374. I don't love it - putting the `__version__` definition above any significant imports is a bit of a 'one weird trick', but it makes getting the version work in more situations without the module having to do anything Flit specific, and I can see it can give better error messages when it fails, which is always good. I see it uses something new in Python 3.5, though - I'm still trying to support 3.4 in flit_core at present. Bumping the minimum Python version for flit_core means leaving any modules which want to support that version stuck on an older flit_core, which is a support headache, so I'm quite slow to do so, but even so, maybe it's time to let 3.4 go now.

I hope you won't mind if I take this chance to run an idea past you for making getting the version number more flexible. This is for discussion, at the moment, rather than implementation, but it's been bouncing around in my head for a while. I'll start with an illustration:

```toml
[tool.flit.version]
method = "import"
module = "foobar.version"
```

I.e. you would specify whether to use static analysis or execution, and which module defines the version. Questions I'm not sure about:

- Import a module or exec a file? Exec-ing allows a `foobar/version.py` to be run without `foobar/__init__.py`, but importing is obviously how modules are normally loaded, and I place quite a high value on interacting with user code in normal, predictable ways.
  - I *could* make both methods available, but I've always tried to avoid adding too many options, so I don't want to do this.
- Should it have an option to look for another name besides `__version__`?
- Should static analysis be able to handle tuples like `version_info = (3, 2, 0, 'beta')`?
- Should I go all in and allow specifying a function to call to retrieve the version number?

I imagine that you'll want maximum flexibility & power on all of these questions, whereas my instinct is to keep things simpler. I'm thinking through how many options I want to enable here. :slightly_smiling_face: 

I'm also thinking that if I do go for this, then the current behaviour of trying static analysis and then falling back to running code will eventually go away (after a suitable deprecation), and you will have to specify explicitly that you want it to import/exec code to get the version number.

flying-sheep:
Sounds great that you are thinking about tackling this!

I’m personally not maintaining things too far back: The slowest Linux distros are on something like 3.7 these days … idk what reasons exist to support 3.4 for things that aren’t build tools. In flit’s case, “this is a build tool and other people still use 3.4” is an entirely valid reason if that’s how it is.

> I imagine that you'll want maximum flexibility & power on all of these questions, whereas my instinct is to keep things simpler.

I love simplicity! Power and simplicity aren’t mutually exclusive: Allowing people to call into whatever and clearly telling them that it’s not a flit bug if things go wrong is maximally powerful and very simple.

I personally want whatever solution allows my packages to robustly tell the truth about their version. And hardcoding a version isn’t the truth, unless you manually bump that version in every commit (which isn’t realistic).

> Questions I'm not sure about: […]

So what use cases do exist?

1. some hardcode `__version__`
2. some call code which depends on `build-system.requires`

The problem with use case 2 is that flit executes the root module before it installs all dependencies, which means that people with use case 2 can’t both use flit and import anything in their root module (unless they use my hack).

To truly support use case 2, we need to give people a way to execute the version-generating code without executing the root module, or install all dependencies before executing the root module (#141)

There’s three ways to run code without (fully) executing the root module:

1. This PR: Try to execute the root module and pick `__version__` out even when an `ImportError` happened
2. Specify an out-of-module script/module to run/import (could be either a third party module or a standalone script living next to the module)
3. Exec-ing a in-module script as if it wasn’t a submodule

PR #390, opened 2021-02-03
Allow `Private :: Do Not Upload` classifier.

orsinium:
Listing "Private :: Do Not Upload" in the classifiers list is a way to protect a package from accidental uploading on PyPI. This is a special case because it is not listed in the trove classifier.

References:

+ Origin: https://twitter.com/di_codes/status/1097220464445931525
+ It is implemented on PyPI side: https://github.com/pypa/warehouse/pull/5440
+ It's not documented yet: https://github.com/pypa/packaging.python.org/issues/643
+ It is supported by twine: https://github.com/pypa/twine/issues/557

P.S. Sorry for PR without making an issue upfront. I think this is the case that isn't hard to implement, so it's better to show than explain :)

### DIFF
--- a/doc/pyproject_toml.rst
+++ b/doc/pyproject_toml.rst
@@ -92,6 +92,8 @@ description-file
   (``.rst``, ``.md`` or ``.txt``).
 classifiers
   A list of `Trove classifiers <https://pypi.python.org/pypi?%3Aaction=list_classifiers>`_.
+  Add ``Private :: Do Not Upload`` into the list to prevent a private package
+  from uploading on PyPI by accident.
 requires-python
   A version specifier for the versions of Python this requires, e.g. ``~=3.3`` or
   ``>=3.3,<4`` which are equivalents.
--- a/flit/validate.py
+++ b/flit/validate.py
@@ -13,6 +13,12 @@ from .vendorized.readme.rst import render
 
 log = logging.getLogger(__name__)
 
+CUSTOM_CLASSIFIERS = frozenset({
+    # https://github.com/pypa/warehouse/pull/5440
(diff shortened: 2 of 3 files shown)


takluyver:
Thanks, this is an interesting one. The same would work with any invalid classifier, there's nothing particularly special about the name `Private :: Do Not Upload`. But of course, Flit is deliberately checking if classifiers are valid.

I see from your link that PyPI is essentially promising not to create any `Private ::` classifier. Maybe Flit should allow any classifiers in that namespace, on the grounds that if you've done that, it's probably deliberate rather than a mistake. Do you know if any other components of the ecosystem have made changes around this idea?

orsinium:
> Maybe Flit should allow any classifiers in that namespace

That would work but I'm for keeping things more strict, there is no reason to have anything different like `Private :: Some Explanation Why`, additional context can be provided in a comment next to the classifier.

> Do you know if any other components of the ecosystem have made changes around this idea?

I don't know any other that checks the classifier on the client-side, TBH. AFAIK, `setuptools`, `poetry`, `twine`, and `wheel` don't. In `dephell`, I had only a [limited classifier validation](https://github.com/dephell/dephell/blob/master/dephell/commands/project_validate.py#L73) as well.

What about usage, GitHub search pops a bit of results:
https://github.com/search?q=%22Private+%3A%3A+Do+Not+Upload%22&type=code
Of course, there won't be much because, well, it's strange to see "private" libraries in public repos :)

orsinium:
Is there anything I can do to have it merged into the upstream?

takluyver:
No, I think that's OK, thanks.

PR #400, opened 2021-03-20
PEP 660 support

sbidoul:
This is a very rough proof of concept for `build_wheel_for_editable`, as described in https://discuss.python.org/t/standardising-editable-mode-installs-runtime-layout-not-hooks/4098/55 and prototyped in https://github.com/pypa/pip/pull/8212. 

It works by adding a `.pth` file in the transient wheel, and should therefore behave identically as `flit install --pth-file`.

### DIFF
--- a/flit_core/flit_core/wheel.py
+++ b/flit_core/flit_core/wheel.py
@@ -146,6 +146,10 @@ class WheelBuilder:
             rel_path = osp.relpath(full_path, source_dir)
             self._add_file(full_path, rel_path)
 
+    def add_pth(self):
+        with self._write_to_zip(self.module.name + ".pth") as f:
+            f.write(str(self.module.source_dir))
+
     def write_metadata(self):
         log.info('Writing metadata files')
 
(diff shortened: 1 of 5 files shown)


sbidoul:
Updated to match the latest PEP 660 version.

flit_core/flit_core/wheel.py
```diff
@@ -189,6 +189,14 @@ def copy_module(self):
             rel_path = osp.relpath(full_path, source_dir)
             self._add_file(full_path, rel_path)
 
+    def add_pth(self):
+        pth_target = osp.dirname(osp.abspath(self.module.path))
+        pth_file = self.module.path.name + ".pth"
+        with tempfile.NamedTemporaryFile("w") as f:
+            f.write(pth_target)
+            f.flush()
+            self._add_file(f.name, pth_file)
```
takluyver, line 198: You could use `with self._write_to_zip(pth_file):` here, I think - this is what we use elsewhere for generated files.

takluyver:
Thanks @sbidoul - this looks reasonable enough to me, if you'd like to add a test or two.

flit_core/flit_core/wheel.py
```diff
@@ -189,6 +189,14 @@ def copy_module(self):
             rel_path = osp.relpath(full_path, source_dir)
             self._add_file(full_path, rel_path)
 
+    def add_pth(self):
+        pth_target = osp.dirname(osp.abspath(self.module.path))
```
takluyver, line 193: We've got `self.module.source_dir` as a shortcut for this directory - you just need to `str()` it.

sbidoul:
Thanks for the review. I should come back to this later this month.

sbidoul:
@takluyver this should now be ready.

I'm not sure how to test `buildapi.build_editable` as existing tests for buildapi rely on the filt command and I don't think we should expose the editable wheel format there.

takluyver:
Thanks, I'll take a look. :-)

I'm not sure what you mean about tests for buildapi relying on the `flit` command - this one looks like it's calling buildapi directly, for instance:

https://github.com/takluyver/flit/blob/734ab12fc5bfadf7e37dc303d44fd517f7ebda1b/flit_core/flit_core/tests/test_buildapi.py#L44-L49

takluyver:
Nice and simple, thank-you. I think if you can add a simple test to exercise the function exposed in buildapi, this should be fine.

sbidoul:
@takluyver I somehow missed the buildapi tests in flit_core. Proof I needed holidays, I guess.

I now added the missing tests.

sbidoul:
@takluyver FYI PEP 660 has been merged in pip and will be part of the 21.3 release in October. \o/

takluyver:
Thanks, and congratulations on getting the PEP you worked on accepted & implemented!

sbidoul:
Thanks, it was quite a ride, I should say :)

PR #410, opened 2021-05-01
Reuse wheel metadata in sdist

### LINKED ISSUE #409: PKG-INFO values mangled when using PEP 621 metadata
In `PKG-INFO`, flit 3.2.0 has:

```
Home-page: None
Author: None
Author-email: Thomas Kluyver <thomas@kluyver.me.uk>
```

3.1.0 has:

```
Home-page: https://github.com/takluyver/flit
Author: Thomas Kluyver
Author-email: thomas@kluyver.me.uk
```

The 3.2.0 wheel `METADATA` do not have have a `Home-page` nor an `Author` and the `Author-email` is `Thomas Kluyver <thomas@kluyver.me.uk>`, like in the sdist.

When uploading an sdist with a `Home-Page` value of `None` using twine, the legacy upload API rejects it with:

```
'None' is an invalid value for Home-Page.
```

layday:
I was not able to tell if there is a downside to using the same
metadata as the wheel but a cursory investigation revealed
that it's what Poetry does.  setuptools omits the requirements
but its PKG-INFO appears to be otherwise identical to the wheel's
METADATA file.  flit appears to be a bit of an outlier in creating
a barebones PKG-INFO file.

Closes #409.

### DIFF
--- a/flit_core/flit_core/sdist.py
+++ b/flit_core/flit_core/sdist.py
@@ -15,17 +15,6 @@ from . import common
 log = logging.getLogger(__name__)
 
 
-PKG_INFO = """\
-Metadata-Version: 1.1
-Name: {name}
-Version: {version}
-Summary: {summary}
-Home-page: {home_page}
-Author: {author}
-Author-email: {author_email}
-"""
-
-
 def clean_tarinfo(ti, mtime=None):
     """Clean metadata from a TarInfo object to make it more reproducible.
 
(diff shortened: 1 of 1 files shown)


dholth:
Seems like a good idea to me.

henryiii:
This looks much better to me, and would fix twine warnings too.

takluyver:
Thanks! Yes, as far as I know, PKG-INFO is the same format as METADATA.

I think this separate PKG-INFO template was just a holdover from when I rather grudgingly implemented sdist generation, after initially arguing that wheels-only was fine for pure Python packages.

Issue #418, opened 2021-07-06
FileNotFoundError when running build or publish


fanoway:
There are no errors when using the init or install commands bunning build or install returns the following error:

```C:\Users\User Name\Downloads\entrypoint>pipenv run flit build
Traceback (most recent call last):
  File "c:\users\user name\appdata\local\programs\python\python39\lib\runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "c:\users\user name\appdata\local\programs\python\python39\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\eser name\.virtualenvs\entrypoint-RYMZNUJV\Scripts\flit.exe\__main__.py", line 7, in <module>
  File "c:\users\user name\.virtualenvs\entrypoint-rymznujv\lib\site-packages\flit\__init__.py", line 154, in main
    main(args.ini_file, formats=set(args.format or []),
  File "c:\users\user name\.virtualenvs\entrypoint-rymznujv\lib\site-packages\flit\build.py", line 46, in main
    sdist_file = sb.build(dist_dir, gen_setup_py=gen_setup_py)
  File "c:\users\user name\.virtualenvs\entrypoint-rymznujv\lib\site-packages\flit_core\sdist.py", line 182, in build
    files_to_add = self.apply_includes_excludes(self.select_files())
  File "c:\users\user name\.virtualenvs\entrypoint-rymznujv\lib\site-packages\flit\sdist.py", line 146, in select_files
    untracked_deleted = vcs_mod.list_untracked_deleted_files(self.cfgdir)
  File "c:\users\user name\.virtualenvs\entrypoint-rymznujv\lib\site-packages\flit\vcs\git.py", line 12, in list_untracked_deleted_files
    outb = check_output(['git', 'ls-files', '--deleted', '--others',
  File "c:\users\user name\appdata\local\programs\python\python39\lib\subprocess.py", line 424, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "c:\users\user name\appdata\local\programs\python\python39\lib\subprocess.py", line 505, in run
    with Popen(*popenargs, **kwargs) as process:
  File "c:\users\user name\appdata\local\programs\python\python39\lib\subprocess.py", line 951, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
[...]

PR #419, opened 2021-07-06
Fix to issue #418

fanoway:
See example error when git is stored in .git folder

https://github.com/takluyver/flit/issues/418

### DIFF
--- a/flit/vcs/git.py
+++ b/flit/vcs/git.py
@@ -1,15 +1,27 @@
 import os
 from subprocess import check_output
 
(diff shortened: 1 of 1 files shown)


takluyver:
What is your `git` command? On my system, `git ls-files` works, and `git .git ls-files` gives me an error:

```
git: '.git' is not a git command. See 'git --help'.
```

fanoway:
I was on  a fresh windows install. Restarting my pc seemed to fix this. Thanks

Issue #426, opened 2021-07-30
Support alternativ location for .pypirc (for Azure Devops Pipelines)


decatur:
Use case is to let `flit publish` on Azure Devops Pipelines.
All files created with a pipeline should reside in the Pipeline.Workspace, which typically points to something like `/home/vsts/work/1`.

As a helper Task, there is [TwineAuthenticateV1](https://github.com/microsoft/azure-pipelines-tasks/tree/master/Tasks/TwineAuthenticateV1), which creates a .pypirc in $PYPIRC_PATH. The latter defaulting to something like `/home/vsts/work/_temp/twineAuthenticate/zhVvTB/.pypirc`

Now ~ expands to `/home/vsts`, which is NOT documented to be writeable by pipelines.

So flit should support altering the location of the .pypirc file (like --config-file for twine)

Workaround: Move .pypirc to `/home/vsts`, which currently works.

[Typical pipeline definition with workaround](https://github.com/takluyver/flit/files/6907534/flit_issue.txt)

takluyver:
The obvious way to pass credentials in a CI job is with environment variables, avoiding the need for a file entirely:

https://flit.readthedocs.io/en/latest/upload.html#using-environment-variables

decatur:
No such luck.

Azure DevOps has a layered approach to security.
As a developer or distribution publisher, you do not own nor know the credentials to publish to an index.
These credentials are provided by the CI cloud administrator in the form of a service connection.
Which in turn will produce a .pypirc file.
See the [example](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/package/twine-authenticate?view=azure-devops#examples).

I figured it is easier to get `flit` patched than to move Azure DevOps 😉

davidrft:
I think supporting an argument to specify the location of the pypirc file just as twine does would be the ideal.

davidrft:
I would be happy to try and contribute if we agree on this. Thanks!

decatur:
Suits me fine, i.e.

> --config-file CONFIG_FILE 
>     The .pypirc config file to use.

takluyver:
OK, go for it. I might call the option something like `--pypirc` instead of `--config-file`, because 'config file' seems a bit generic, but other than that.

davidrft:
I worked on the necessary changes to achieve this on #434. Could you please review it once you have some time? Thanks!

PR #434, opened 2021-08-16
Add custom pypirc support

davidrft:
This PR adds support for handling a custom path for the `.pypirc` file. Previously the file was assumed to always be in `~/.pypirc`, which can be an issue depending on the build environment that is being used as stated in #426.

The path can be passed through the `--pypirc` argument, as such:
```sh
flit publish --pypirc .pypirc
```

A unit test was added to ensure the parsing is correctly done when the custom path is specified.

### DIFF
--- a/doc/cmdline.rst
+++ b/doc/cmdline.rst
@@ -73,7 +73,7 @@ or another repository.
 
 .. option:: --pypirc <pypirc>
 
-   The .pypirc config file to be used. DEFAULT = "~/.pypirc"
+   The .pypirc config file to be used. The default is ``~/.pypirc``.
 
 .. seealso:: :doc:`upload`
 
--- a/doc/cmdline.rst
+++ b/doc/cmdline.rst
@@ -71,6 +71,10 @@ or another repository.
    Name of a repository to upload packages to. Should match a section in
    ``~/.pypirc``. The default is ``pypi``.
 
+.. option:: --pypirc <pypirc>
+
+   The .pypirc config file to be used. DEFAULT = "~/.pypirc"
+
 .. seealso:: :doc:`upload`
 
 .. _install_cmd:
--- a/doc/history.rst
+++ b/doc/history.rst
@@ -316,8 +316,8 @@ Version 0.11
 
   - If you have the `keyring <https://github.com/jaraco/keyring>`_ package
     installed, flit can use it to store your password, rather than keeping it
-    in plain text in ``~/.pypirc``.
-  - If ``~/.pypirc`` does not already exist, and you are prompted for your
+    in plain text in the .pypirc file.
+  - If the .pypirc file does not already exist, and you are prompted for your
     username, flit will write it into that file.
   - You can provide the information as environment variables:
     :envvar:`FLIT_USERNAME`, :envvar:`FLIT_PASSWORD` and :envvar:`FLIT_INDEX_URL`.
--- a/doc/history.rst
+++ b/doc/history.rst
@@ -316,8 +316,8 @@ Version 0.11
 
   - If you have the `keyring <https://github.com/jaraco/keyring>`_ package
     installed, flit can use it to store your password, rather than keeping it
-    in plain text in the .pypirc file.
-  - If the .pypirc file does not already exist, and you are prompted for your
+    in plain text in ``~/.pypirc``.
+  - If ``~/.pypirc`` does not already exist, and you are prompted for your
     username, flit will write it into that file.
   - You can provide the information as environment variables:
     :envvar:`FLIT_USERNAME`, :envvar:`FLIT_PASSWORD` and :envvar:`FLIT_INDEX_URL`.
--- a/flit/__init__.py
+++ b/flit/__init__.py
@@ -116,6 +116,10 @@ def main(argv=None):
              )
     )
 
+    parser_publish.add_argument('--pypirc', default="~/.pypirc",
+        help="Path for .pypirc file."
+    )
+
     parser_publish.add_argument('--repository',
         help="Name of the repository to upload to (must be in ~/.pypirc)"
     )
@@ -184,7 +188,7 @@ def main(argv=None):
             log.warning("Passing --repository before the 'upload' subcommand is deprecated: pass it after")
         repository = args.repository or args.deprecated_repository
         from .upload import main
-        main(args.ini_file, repository, formats=set(args.format or []),
+        main(args.ini_file, repository, args.pypirc, formats=set(args.format or []),
                 gen_setup_py=gen_setup_py())
 
     elif args.subcmd == 'install':
--- a/flit/__init__.py
+++ b/flit/__init__.py
@@ -117,11 +117,11 @@ def main(argv=None):
     )
 
     parser_publish.add_argument('--pypirc', default="~/.pypirc",
-        help="Path for .pypirc file."
+        help="The .pypirc config file to be used. DEFAULT = \"~/.pypirc\""
     )
 
     parser_publish.add_argument('--repository',
-        help="Name of the repository to upload to (must be in ~/.pypirc)"
+        help="Name of the repository to upload to (must be in the specified .pypirc file)"
     )
 
     # flit install --------------------------------------------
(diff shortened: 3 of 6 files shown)


doc/cmdline.rst
```diff
@@ -71,6 +71,10 @@ or another repository.
    Name of a repository to upload packages to. Should match a section in
    ``~/.pypirc``. The default is ``pypi``.
 
+.. option:: --pypirc <pypirc>
+
+   The .pypirc config file to be used. DEFAULT = "~/.pypirc"
```
takluyver, line 76:
```suggestion
   The .pypirc config file to be used. The default is ``~/.pypirc``.
```

flit/upload.py
```diff
@@ -59,13 +59,13 @@ def get_repositories(file="~/.pypirc"):
     return repos
 
 
-def get_repository(name=None, cfg_file="~/.pypirc"):
```
takluyver, line 62: I'm a bit surprised by the random extra changes - reordering and renaming parameters, removing default values. It doesn't really matter, and as long as this is all working, it's OK, but it's a good idea in general to keep your changes focused on the specific thing you're trying to change. More changes means it's more work to review. :wink:

flit/upload.py
```diff
@@ -85,7 +85,9 @@ def get_repository(name=None, cfg_file="~/.pypirc"):
     3. keyring
     4. Terminal prompt (store to keyring if available)
     """
-    repos_cfg = get_repositories(cfg_file)
+    log.info(pypirc_path)
+    repos_cfg = get_repositories(pypirc_path)
+    log.info(repos_cfg)
```
takluyver, line 90:
```suggestion
    log.debug("Loading repositories config from %r", pypirc_path)
    repos_cfg = get_repositories(pypirc_path)
```

Log messages at info level and above are visible to the user by default, so I like to be fairly minimal with them. Debug level is only shown if you add `--debug`, which seems about right for this.

`repos_cfg` may contain passwords, so it's probably best that we don't log that at all! You don't want it to give away your password in a CI log.

flit/upload.py
```diff
@@ -237,10 +239,10 @@ def upload_file(file:Path, metadata:Metadata, repo):
     resp.raise_for_status()
 
 
-def do_upload(file:Path, metadata:Metadata, repo_name=None):
+def do_upload(file:Path, metadata:Metadata, pypirc_path:Path, repo_name=None):
```
takluyver, line 242:
```suggestion
def do_upload(file:Path, metadata:Metadata, pypirc_path, repo_name=None):
```

I don't see anything converting the string to a Path object, and I'm not sure `get_repositories` would work if it was.

flit/upload.py
```diff
@@ -124,7 +126,7 @@ def get_repository(name=None, cfg_file="~/.pypirc"):
 
     return repo
 
-def write_pypirc(repo, file="~/.pypirc"):
+def write_pypirc(repo, file):
```
takluyver, line 129: You've changed the signature here, but not updated where the function is called (in `get_repository`).

flit/__init__.py
```diff
@@ -116,8 +116,12 @@ def main(argv=None):
              )
     )
 
+    parser_publish.add_argument('--pypirc', default="~/.pypirc",
```
takluyver, line 119: I think we should handle it slightly differently if this is explicitly specified - it should check that the specified file exists, because if you've specified an exact path and it doesn't exist, that's probably a mistake.

That would mean not setting an explicit default here, so we can distinguish whether it's given at the command line or not.

takluyver (commented):
Thanks - this is mostly OK, but needs a few adjustments.

flit/upload.py
davidrft: Sorry about that. 

1. I renamed the parameters to keep it mostly consistent with what we see in the interface.
1. Removed default values so they can have an only source of truth to improve maintainability in case we need to change the default value and/or debug.
1. Reordered some function in the test so it could be used in a previous test.

Please let me know if you don't think any of those changes are necessary and I would be glad to revert! :)

flit/upload.py
davidrft: Thanks, good catch. You are completely right; this was left over from my debugging, I'll change that.

flit/__init__.py
davidrft: Thanks, that's a good suggestion! I removed the default so we can use None as a sentinel and checked the file existence. Because of that I needed to patch `os.path.isfile` in the unit tests. Let me know your thoughts, thanks!

flit/upload.py
takluyver: I kind of like default values even if it involves some repetition, because it's an easy way to see what sort of value is expected, without having to go and look at other layers of code. And I don't think the default is likely to change. :smiley: 

E.g. here, I can see that `cfg_file` is a string path, that it needs `~` to be expanded to the home directory, and that it refers to a .pypirc config file rather than some other type of config. That's a lot of useful information - more than you're likely to have in type annotations - for a very small bit of extra code.

So maybe we can put the default values back. More generally, I'd say "if it ain't broke, don't fix it" :wink: , but there's no need to revert all the changes now.

davidrft: Got it, no worries! I can add the default values back later today if you wish.

tests/test_upload.py
```diff
@@ -87,7 +92,53 @@ def get_password(service_name, username):
...
+
+[test123]
+repository: {pypirc3_repo}
+username: {pypirc3_user}
+password: {pypirc3_pass}
+"""
+
+
+def test_upload_pypirc_file(copy_sample):
+    with patch('os.path.isfile', return_value=True), \
```
takluyver, line 117: I don't particularly like having to mock this out. I think we should decide whether the functions accept file-like objects, and make that work without having to mock anything, or take only file paths, and test using temp files.

I'd say use temp files. Pytest makes that pretty convenient, and the code under test shouldn't really have special additional code paths just for the tests.

doc/history.rst
```diff
@@ -316,8 +316,8 @@ Version 0.11
 
   - If you have the `keyring <https://github.com/jaraco/keyring>`_ package
     installed, flit can use it to store your password, rather than keeping it
-    in plain text in ``~/.pypirc``.
-  - If ``~/.pypirc`` does not already exist, and you are prompted for your
+    in plain text in the .pypirc file.
+  - If the .pypirc file does not already exist, and you are prompted for your
```
takluyver, line 320: I might ask you to revert this bit - it's not normally necessary to modify release notes from several years ago, unless there's a broken link or something.

davidrft: Sure! Sorry, I thought this was documentation. Will revert.

tests/test_upload.py
davidrft: Ok, will change the approach to using temp files.

doc/history.rst
takluyver: No problem! It is part of the docs, but it's specifically saying what changed from one release to the next, and this bit is about the 0.11 release a few years ago.

tests/test_upload.py
davidrft: Done, please let me know what you think. Thanks!

tests/test_upload.py
```diff
@@ -38,19 +32,43 @@ def test_upload(copy_sample):
...
 
+@contextmanager
+def temp_pypirc(content):
+    try:
+        temp_file = NamedTemporaryFile("w+", delete=False)
+        temp_file.write(content)
+        temp_file.close()
+        yield temp_file.name
+    finally:
+        os.unlink(temp_file.name)
```
takluyver, line 43: Pytest has a handy shortcut for this - if you use a `tmp_path` fixture in a test, it creates a temporary directory, you can write whatever you want there, and it will all be cleaned up when the test runs. [Docs here](https://docs.pytest.org/en/latest/how-to/tmp_path.html#tmp-path)

I'm happy with this, the fixture is just a neat thing to know about. :slightly_smiling_face:

takluyver:
Thanks!

tests/test_upload.py
davidrft: Thanks! Good to learn about that :)

PR #441, opened 2021-08-28
Document how to bootstrap flit_core

takluyver:
Mainly for people putting Python packages into other packaging systems.

### DIFF
--- a/doc/bootstrap.rst
+++ b/doc/bootstrap.rst
@@ -1,0 +1,35 @@
+Bootstrapping
+=============
+
+Flit is itself packaged using Flit, as are some foundational packaging tools
+such as ``pep517``. So where can you start if you need to install everything
+from source?
+
+.. note::
+
+   For most users, ``pip`` handles all this automatically. You should only need
+   to deal with this if you're building things entirely from scratch, such as
+   putting Python packages into another package format.
+
+The key piece is ``flit_core``. This is a package which can build itself using
+nothing except Python and the standard library. From an unpacked source archive,
+you can run ``python build_dists.py``, of which the crucial part is:
+
+    from flit_core import build_thyself
+    whl_fname = build_thyself.build_wheel('dist/')
+    print(os.path.join('dist', whl_fname))
+
+This produces a ``.whl`` wheel file, which you can unzip into your
+``site-packages`` folder (or equivalent) to make ``flit_core`` available for
+building other packages. (You could also just copy ``flit_core`` from the
+source directory, but without the ``.dist-info`` folder, tools like pip won't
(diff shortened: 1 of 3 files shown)


nanonyme:
This is really awesome, especially if we eventually move towards a situation that things like setuptools require PEP517 to build. (currently it self-bootstraps) The only thing I'm a bit concerned about is that unpacking wheels seems not a really straightforward operation when it comes to entrypoints. There is no processor for those except for pip as far as I've understood and that makes the bootstrapping cycle more complex.

nanonyme:
I was just for PoC trying to see if I can follow this chain. It looks like flit stage 1 -> tomli -> flit -> pep517 works but then build requires setuptools and wheel (wheel also requires setuptools). Setuptools in turn can be built either through setup.py or through build/pip. But pip requires setuptools to be built. There seem to be cycles in this bootstrapping chain. Did I miss something?

It would at least break the chain if build and wheel were converted to be built with flit. Wheel does not even have a pyproject.toml at this point.

nanonyme:
FWIW I wrote minimal PEP517 bootstrapper tool (https://gitlab.com/freedesktop-sdk/pep517_install/-/tree/main) to PoC this chain. Indeed, looks like only wheel is problematic in the chain. It would help if it was built with flit.

takluyver:
Thanks for trying it out! I would imagine that setuptools is meant to self-bootstrap in a vaguely similar way to flit-core, but right now it looks like there's a cycle between setuptools & wheel in a fully PEP-517 world. You can presumably get round that by using the 'old way' `setup.py install` to install setuptools. It's up to them whether that's just what you have to do, or whether there will be a solution for PEP 517 bootstrapping.

Packaging `wheel` with flit_core would also break the cycle. I always try not to push Flit on anyone, though.

> unpacking wheels seems not a really straightforward operation when it comes to entrypoints. There is no processor for those except for pip as far as I've understood...

There's the [installer](https://github.com/pradyunsg/installer/) project. It's still quite new & rough, but this is meant to be the low-level tool for this purpose (in partnership with `build`).

I hope that all of the packages you need to get build & installer can be installed as plain libraries - i.e. just unzip the wheel, no entry points required. Some of them may have non-essential entry points, though.

nanonyme:
I see. So then the chain would probably be flit_core->tomli->installer->flit to get entrypoints fully set up. (flit itself even has one)

nanonyme:
I'll have another round of testing with that set of tooling at some point.

takluyver:
Yup. The `flit` package isn't especially important for bootstrapping - flit_core is what really builds packages, `flit` is mainly handy developer tooling around that.

Issue #345, opened 2020-05-21
Build error with filenames containing unicode characters


bpabel:
I'm trying to build a project that has files that contain unicode characters in their filename.  I can build the wheel successfully, but building the sdist results in the following error.

<img width="351" alt="Screen Shot 2020-05-20 at 11 34 07 PM" src="https://user-images.githubusercontent.com/622914/82530788-ee5c2200-9af2-11ea-9f33-9998f6422a59.png">

```
$ flit build
Found 6 files tracked in git                                                                                                                                   I-flit.sdist
Traceback (most recent call last):
  File "/Users/brendan/.local/bin/flit", line 8, in <module>
    sys.exit(main())
  File "/Users/brendan/.local/pipx/venvs/flit/lib/python3.8/site-packages/flit/__init__.py", line 170, in main
    main(args.ini_file, formats=set(args.format or []),
  File "/Users/brendan/.local/pipx/venvs/flit/lib/python3.8/site-packages/flit/build.py", line 46, in main
    sdist_file = sb.build(dist_dir, gen_setup_py=gen_setup_py)
  File "/Users/brendan/.local/pipx/venvs/flit/lib/python3.8/site-packages/flit/sdist.py", line 223, in build
    return Path(super().build(str(target_dir), gen_setup_py=gen_setup_py))
  File "/Users/brendan/.local/pipx/venvs/flit/lib/python3.8/site-packages/flit_core/sdist.py", line 185, in build
    ti = tf.gettarinfo(path, arcname=pjoin(self.dir_name, relpath))
  File "/Library/Frameworks/Python.framework/Versions/3.8/lib/python3.8/tarfile.py", line 1830, in gettarinfo
    statres = os.lstat(name)
FileNotFoundError: [Errno 2] No such file or directory: '"tests/No\\303\\253l.txt"'
```

takluyver:
Thanks. It looks like `git ls-files` quotes the filename for us when it's called here:

https://github.com/takluyver/flit/blob/cfd71d81be95300266c6e989cf2d1bf57aa637fc/flit/vcs/git.py#L6-L9

I see it has a `-z` option to give the raw filenames delimited by null bytes. That's probably what we need.

bpabel:
@takluyver You were exactly right, thanks!  Changing that fixed the issue (see #346 )

PR #445, opened 2021-09-29
add test using git

mawillcockson:
This adds a few pytest fixtures that set up a git repository using `git`.

I used [this stackoverflow answer](https://stackoverflow.com/a/43882168) to isolate git using these environment variable:

- `GIT_CONFIG_GLOBAL`
- `GIT_CONFIG_NOSYSTEM`
- `HOME`
- `GIT_AUTHOR_EMAIL`
- `GIT_AUTHOR_NAME`
- `GIT_AUTHOR_DATE`
- `GIT_COMMITTER_EMAIL`
- `GIT_COMMITTER_NAME`
- `GIT_COMMITTER_DATE`

Instead of using `GIT_DIR` and `GIT_WORK_TREE`, I added another fixture that forms a closure around git, and calls it's `-C` command-line option, pointing to the temporary repository.

I'm not familiar with testing git, so I'm not sure if these measures are sufficient.

The `tmp_git_repo` fixture initializes the temporary repository with an empty commit. This isn't strictly necessary, it just makes some of the possible tests require less setup.

The `tmp_project` fixture copies the sample module from [`./tests/samples/module1_toml`](https://github.com/takluyver/flit/tree/87162fdeefb0a8d34a22193bf7f6c28828563e1d/tests/samples/module1_toml) to the repository and commits the files.

A test is added for #345 as an example of how to use the new fixtures.

I wanted to be able to run `tox` and `pytest` targeting or excluding only the tests that depend on `git` being installed:

```sh
tox -- -m needgit
```

or skipped:

```sh
tox -- -m "not needgit"
```

I kept getting confused myself, so both `needgit` and `needsgit` work.

Currently this is selected based on the test function name including `needgit` or `needsgit`, and I can understand if that's a bit too magical.

The tests are skipped if `shutil.which("git")` returns nothing, though I did not test this.

---

## Possible changes

Would it make more sense to have `tmp_project` behave similarly to the existing `copy_sample` fixture?

It could take as an argument the sample module and return a [`contextmanager`](https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager)-wrapped iterator that copies the named sample module files into the git repository and commits the changes:

```python
def test_example(copy_to_repo):
    with copy_to_repo("module1_toml") as repo:
        ...
```

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,3 +1,4 @@
+import shlex
 from pathlib import Path
 from shutil import copy, copytree, which
 from subprocess import check_output
@@ -39,7 +40,7 @@ def copy_sample(tmp_path):
 
 def git(repo: Path, command: "Union[List[str], str]") -> bytes:
     if isinstance(command, str):
-        args = command.split()
+        args = shlex.split(command)
     else:
         args = command
 
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -1,4 +1,5 @@
 import shlex
+from functools import wraps
 from pathlib import Path
 from shutil import copy, copytree, which
 from subprocess import check_output
@@ -8,7 +9,9 @@ from unittest.mock import patch
 import pytest
 
 if TYPE_CHECKING:
-    from typing import Iterator, List, Union
+    from typing import Callable, Iterator, List, Union
+
+    GitCmd = Callable[Union[str, List[str]], bytes]
 
 samples_dir = Path(__file__).parent / "samples"
 
@@ -38,14 +41,14 @@ def copy_sample(tmp_path):
     return copy
 
 
-def git(repo: Path, command: "Union[List[str], str]") -> bytes:
+def git_cmd(repository_path: Path, command: "Union[List[str], str]") -> bytes:
     if isinstance(command, str):
         args = shlex.split(command)
     else:
         args = command
 
     return check_output(
-        ["git", "-C", str(repo), *args],
+        ["git", "-C", str(repository_path), *args],
     )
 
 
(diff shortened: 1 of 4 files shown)


takluyver:
Thanks! This looks like a good start. I don't know exactly what's needed to isolate git either, so we may as well try it and fix things as they come up. :smile: 

The thing with the test names is indeed a bit more magic than I like. Let's just use an explicit `@pytest.mark.needsgit` marker. I'd also prefer to have just the one name - I can see the potential for mistakes, but it feels kind of messy to solve that by having both. We can use pytest's `--strict-markers` option to check that the markers in the code are all correct.

I think having the fixtures automatically copy one working sample is a decent starting point for now. But I might try to separate the layer for calling git (including all those environment variables to make it consistent) from the layer that prepares a specific repository. We could always refactor that later when we want more flexibility, though.

Lastly, it looks like the tests are failing on Windows (Appveyor CI). If needs be, we can skip it on Windows and only have the new tests on Linux for now, but if possible it would be good to figure it out.

tests/conftest.py
```diff
@@ -1,15 +1,104 @@
...
+    with patch.dict(
+        "os.environ",
+        {
+            # https://git-scm.com/docs/git#Documentation/git.txt-codeGITCONFIGGLOBALcode
+            "GIT_CONFIG_GLOBAL": str(git_global_config),
+            # https://git-scm.com/docs/git#Documentation/git.txt-codeGITCONFIGNOSYSTEMcode
+            "GIT_CONFIG_NOSYSTEM": "true",
+            "HOME": str(tmp_path),
+            # tox by default only passes the PATH environment variable, so
+            # XDG_CONFIG_HOME is already unset
```
takluyver, line 70: If it's easy to do something appropriate with it here, we may as well not make assumptions about how tox is set up. I often run `pytest` directly for testing locally (though I don't have XDG_CONFIG_HOME set).

The comment may be more characters than overriding it. :stuck_out_tongue: 

There's also the [monkeypatch fixture](https://docs.pytest.org/en/latest/how-to/monkeypatch.html) which can be used to temporarily set & remove environment variables. I have no particular preference if you prefer to use `patch.dict`, though.

mawillcockson: It failed when I tried, because tox had already removed it, and I feel silly that I forgot python has try / catch 🤦

Thanks for letting me know about the monkeypatch fixture.

I personally have a preference for using features implemented by the framework, so I implemented both changes in https://github.com/takluyver/flit/pull/445/commits/a47ae163096a63cf003372eedfbd9d2c613dafc4.

mawillcockson:
Thanks for being welcoming to contributions, and for giving very friendly feedback.

> Let's just use an explicit `@pytest.mark.needsgit` marker
> I'd also prefer to have just the one name

✔️ https://github.com/takluyver/flit/pull/445/commits/7430aee7e4f60a76a269e09e234136245e26bb63

I might try to separate the layer for calling git (including all those environment variables to make it consistent) from the layer that prepares a specific repository.

---

> Lastly, it looks like the tests are failing on Windows (Appveyor CI). If needs be, we can skip it on Windows and only have the new tests on Linux for now, but if possible it would be good to figure it out.

It looks like the tests are working already! 😁

Summary of the bug: `git` returns file paths separated by `/` regardless of the platform.

In the test this pr adds ([starting from this line](https://github.com/takluyver/flit/blob/9335bfbf23bf79b05b30c23e909c073a477e7f65/flit/sdist.py#L147)):

- [`flit.vcs.list_untracked_deleted_files()`](https://github.com/takluyver/flit/blob/9335bfbf23bf79b05b30c23e909c073a477e7f65/flit/vcs/git.py#L11-L15) returns `dist/module1-0.1.tar.gz`
- [`flit.sdist.include_path()` checks if this string starts with `dist\`](https://github.com/takluyver/flit/blob/9335bfbf23bf79b05b30c23e909c073a477e7f65/flit/sdist.py#L83)
- A VCSError is thrown because the `dist/` folder is not excluded

I would be happy to open an issue and PR for a quick fix.

My personal preference with Python is to `pathlib`-ify everything dealing with file paths, and I'd be happy to open another PR to do this.

PR #449, opened 2021-10-02
In PEP 660 mode, the pth must point to an absolute directory

sbidoul:


sbidoul:
Without this `pip install -e .` does not do what one expects.

takluyver:
Thanks!

PR #453, opened 2021-10-13
Implement implicit namespace packages

martenlienen:
Since the implementation of PEP621 in flit, we can get namespace packages basically for free. The only required change is not reading or importing the module, when the information (description and version) has already been provided in `pyproject.toml`. This is accomplished with just one `if`. The other places already avoid reading the file when `project.dynamic` is empty in `pyproject.toml`. During `sdist` we also have to just not check for the module file.

Best,
Marten

### DIFF
--- a/flit_core/flit_core/buildapi.py
+++ b/flit_core/flit_core/buildapi.py
@@ -28,8 +28,9 @@ def get_requires_for_build_wheel(config_settings=None):
     want_summary = 'description' in info.dynamic_metadata
     want_version = 'version' in info.dynamic_metadata
 
-    module = Module(info.module, Path.cwd())
-    docstring, version = get_docstring_and_version_via_ast(module)
+    if want_summary or want_version:
+        module = Module(info.module, Path.cwd())
+        docstring, version = get_docstring_and_version_via_ast(module)
 
     if (want_summary and not docstring) or (want_version and not version):
         return info.metadata.get('requires_dist', [])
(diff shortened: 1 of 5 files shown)


martenlienen:
I have used this PR successfully today via
```yaml
[build-system]
requires = ["flit_core @ git+https://github.com/martenlienen/flit.git@namespace-packages#subdirectory=flit_core"]
```
to release [hydra-experiment-sweeper](https://pypi.org/project/hydra-experiment-sweeper), a plugin for [hydra](https://github.com/facebookresearch/hydra) that needs to be packaged as a namespace package.

takluyver:
Thanks @martenlienen, and sorry it's taken me a while to look at this.

Truth be told, I like the old approach of #159 better, where you package something *in* the namespace, rather than the namespace itself. I.e. you would specify:

```
[tool.flit.module]
name = "hydra_plugins.experiment_sweeper_plugin"
```

And then it would work like any other Flit package - with support for dynamic version & description fields extracted from the package itself. I'm not be a big fan of namespace packages myself, but if we're going to support them, I'd rather do so properly than as a half feature.

martenlienen:
Thanks for getting back to namespace packages after all. I agree that the other approach feels nicer. The obvious advantage of my PR is just that it is such a small change. Either way, I am looking forward to getting namespace support in flit eventually!

takluyver:
Thanks! A small diff is certainly nice, but I'd rather not have to explain that certain features aren't available when you're packaging something in a namespace package. :slightly_smiling_face:

PR #467, opened 2021-11-11
add license field to pyproject during init

ucodery:
I noticed myself adding this to pyroject.toml after every new project init so I thought it would be a good addition.


takluyver:
Good idea, thanks!

PR #468, opened 2021-11-14
Namespace package support

### LINKED ISSUE #370: Implicit namespace package support
The implicit namespace package is defined in [PEP 420](https://www.python.org/dev/peps/pep-0420/) as lacking a `__init__.py` in the package. But Flit relies on the file to read in the version and (short) description.

Some ways to support this from the top of my head:

1. Add an alternative way to specify those values. Probably against Flit’s philosophy.
2. Add a field in `pyproject.toml` to instruct Flit to read those values from another file. Implicit namespace packages usually implement a submodule in the parent package, so it’d be possible to have something like `top-level = "foo.bar"` to instruct Flit to load that module instead of the one defined under the `module` key.
3. Like the above, but merge this into `module`. So instead of taking a Python identifier, `module` can now may also be given a dotted name to specify a submodule. This would make Flit import `module` for metadata values (as currently), but package all of the top-level module into the distribution. The downside is there’s no way to package multiple submodules into one namespace package in a distribution, but that’s a niche usage anyway (like Flit chose not to support having multiple packages in a distribution).
4. Continue to require a `__init__.py` in the top-level package, but introduce a flag (e.g. `namespace = true`) to not package that file into the distribution. I don’t like this approach (the module would behave differently depending on whether it’s imported directly from PWD or after `flit install -s`), but it’s the easiest hack.

takluyver:
Rebase of #159. I still want to check through this a bit more before merging, but it's nearly ready.

Closes #370, closes #453


martenlienen:
Thanks a lot for this!

PR #474, opened 2021-11-27
Fix ast version parser for multiple assignments

jameshilliard:
Some projects seem to use multiple assignments which currently only seems to work if `__version__` is the first variable being assigned to.

### DIFF
--- a/flit_core/flit_core/common.py
+++ b/flit_core/flit_core/common.py
@@ -132,15 +132,18 @@ def get_docstring_and_version_via_ast(target):
     for child in node.body:
         # Only use the version from the given module if it's a simple
         # string assignment to __version__
-        is_version_str = (
-                isinstance(child, ast.Assign)
-                and len(child.targets) == 1
-                and isinstance(child.targets[0], ast.Name)
-                and child.targets[0].id == "__version__"
-                and isinstance(child.value, ast.Str)
-        )
-        if is_version_str:
-            version = child.value.s
+        if isinstance(child, ast.Assign):
+            for target in child.targets:
+                is_version_str = (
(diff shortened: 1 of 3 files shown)


jameshilliard:
rebased

joukewitteveen (approved):

flit_core/flit_core/common.py
```diff
@@ -132,15 +132,18 @@ def get_docstring_and_version_via_ast(target):
...
+                is_version_str = (
+                    isinstance(target, ast.Name)
+                    and target.id == "__version__"
+                    and isinstance(child.value, ast.Str)
+                )
+                if is_version_str:
+                    version = child.value.s
+                    break
+            else:
+                continue
```
takluyver, line 146: I think there's a good opportunity to use the `any()` builtin here and make this clearer.


```suggestion
        is_version_str = (
                isinstance(child, ast.Assign)
                and any(
                    isinstance(target, ast.Name)
                    and target.id == "__version__"
                    for target in child.targets
                )
                and isinstance(child.value, ast.Str)
        )
        if is_version_str:
            version = child.value.s
            break
```

flit_core/flit_core/common.py
```diff
@@ -134,14 +134,17 @@ def get_docstring_and_version_via_ast(target):
...
+                    isinstance(target, ast.Name)
+                    and target.id == "__version__"
+                    for target in child.targets
+                )
                 and isinstance(child.value, ast.Str)
         )
         if is_version_str:
             version = child.value.s
             break
+            break
```
takluyver, line 147:
```suggestion
```

PR #475, opened 2021-11-29
Generate a setup.py for flit_core for bootstrapping.

jameshilliard:


takluyver:
Thanks, but no thanks, for the reasons I described in our discussion at https://github.com/takluyver/flit/pull/462#issuecomment-981499124 .

PR #478, opened 2021-12-01
Replace /takluyver/ by /pypa/ where relevant

Carreau:
While GitHub does redirect, it is good practice to fix links.

And the redirect will last only until Thomas decide to fork it again.

-- 

Note that my autoformater was a bit overzealous in replacing some simple to double quotes, I can redo those files if you wish to.


takluyver:
Thanks Matthias!

takluyver:
I just happened to see the branch name in git logs - I don't think I had noticed it before. :rofl:

PR #480, opened 2021-12-01
Update development.rst

jdrumgoole:
Missing a package from the list of packages. Specifically tomli.

### DIFF
--- a/doc/development.rst
+++ b/doc/development.rst
@@ -5,7 +5,7 @@ To get a development installation of Flit itself::
 
     git clone https://github.com/takluyver/flit.git
     cd flit
-    python3 -m pip install docutils requests toml
+    python3 -m pip install docutils requests toml tomli
     python3 bootstrap_dev.py
 
 This links Flit into the current Python environment, so you can make changes


doc/development.rst
```diff
@@ -5,7 +5,7 @@ To get a development installation of Flit itself::
 
     git clone https://github.com/takluyver/flit.git
     cd flit
-    python3 -m pip install docutils requests toml
+    python3 -m pip install docutils requests toml tomli
```
hukkin, line 8:
```suggestion
    python3 -m pip install docutils requests tomli tomli-w
```

jdrumgoole: My install didn't complain about tomli-w being missing

hukkin: Ah I see, I think that's because the `bootstrap_dev.py` installs `flit` dependencies if they aren't installed already. So this line only strictly needs to list dependencies that are used by whatever parts of the codebase `bootstrap_dev.py` runs.

I still think this line may be easier to grasp and maintain if it simply installs all flit dependencies.

Regardless, the `toml` dependency should no longer be needed here.

Carreau: I think there are 2 reasons one of them being explained on bootstrap.rst. It's that `tomli` is itself packaged with flit, so to bootstrap on packaging system like conda-forge you want to be able to maybe call `bootstrap_dev`, before tomli/tomli-w are themselves packaged.

Not sure.

takluyver: `bootstrap_dev.py` is really meant for development, rather than downstream packaging - it does an editable install of both flit_core & flit. It might make sense to add another script to help with bootstrapping `flit_core`.

takluyver:
I think it's OK to leave `tomli-w` out of this list - it should get installed anyway as a dependency of Flit. But let's take `toml` out of the list, because tomli replaced that. :slightly_smiling_face:

joukewitteveen:
Did #573 change anything here? On modern systems, `tomli` shouldn't be needed anymore, yet `tomli-w` will be.

takluyver:
I don't think any TOML libraries should need to be manually installed any more. `tomli` is bundled in flit_core, and `tomli_w` is only used for a couple of pieces like `flit init`, so Flit should be able to install itself without that. `tomli_w` and any other missing dependencies will still get installed - we only need to pre-install enough to handle installation.

joukewitteveen:
I guess that means this PR can be closed and replaced by one that removes `toml` from line 8 of `development.rst` (the same line that this PR tried to add `tomli` to).

takluyver:
Yup, that sounds right. Feel free to make that PR. :slightly_smiling_face:

PR #485, opened 2021-12-07
flit-core: add comment explaining where deps are listed

### LINKED ISSUE #482: flit-core: pyproject.toml missing dependencies on toml and tomli

adamjstewart:
Closes #482 

@takluyver @pradyunsg

### DIFF
--- a/flit_core/pyproject.toml
+++ b/flit_core/pyproject.toml
@@ -1,6 +1,7 @@
 [build-system]
-# Dependencies listed in flit_core/build_thyself.py
-# https://github.com/pypa/flit/issues/482#issuecomment-987769343
 requires = []
 build-backend = "flit_core.build_thyself"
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks! Just to be clear, the `requires` list under `[build-system]` is a list of build dependencies, which should be empty - `flit_core` builds itself with no dependencies. It's the `[project]` table, which would contain runtime dependencies, which is completely absent.

PR #492, opened 2021-12-23
Vendor tomli 1.2.3 in flit_core

takluyver:
This avoids the dependency cycle, and also means we don't have to special case flit_core building itself, because we can rely on having a TOML parser.

I've left a note on how it can be unbundled, but of course if people do that, they will need to deal with the dependency cycle some other way. One option would be to get flit_core and tomli installed before unbundling, another would be to make them both available without installing to use for a special bootstrapping step (like [python-bootstrap](https://github.com/FFY00/python-bootstrap) does).

This should be a special case, and hopefully can go away again once there's a TOML parser in the Python standard library.

Closes #483 (I wanted to avoid the extra layers of tooling & automation that PR proposes, as this is just one package)


flit_core/flit_core/config.py
```diff
@@ -5,9 +5,9 @@
 import os
 import os.path as osp
 from pathlib import Path
-import tomli
 import re
 
+from .vendor import tomli
```
pradyunsg, line 10: Do you want to bother with underscore prefixes?

pradyunsg (commented):
Seems fine to me!

I'm wary of how downstream debundlers will mess with things, but... eh, we can get to that when we get there.

flit_core/flit_core/config.py
mgorny:
```suggestion
try:
    from .vendor import tomli
except ImportError:
    import tomli
```

How about this? It would make unbundling a little bit easier.

You could even do it the opposite way — i.e. prevent the "newest" tomli available on the system, and fall back to vendoring only when really necessary.

mgorny: Sorry, I have no clue what distracted me at the time. I meant "prefer".

pradyunsg: If someone is debundling (i.e. modifying the package), they can also do the extra work of updating references. There’s no reason to bake support in unless someone with that use case comes and asks for it; and provides a clear justification.

gaborbernat (approved):

flit_core/flit_core/config.py
takluyver: I'm inclined to leave it as is, because catching the error and falling back could also mask a real error (e.g. if tomli gained a dependency and I bundled it without noticing). Unlikely, but as @pradyunsg says, it's also not much work for someone who wants to unbundle it.

jameshilliard:
> This should be a special case, and hopefully can go away again once there's a TOML parser in the Python standard library.

Just a FYI, the vendored `tomli` should not be removed until a TOML parser is available in python stdlib for all versions of python flit intends to support, although once available in some versions it can be used opportunistically instead of the vendored version at runtime of course.

Looks good. :)

takluyver:
> the vendored tomli should not be removed until a TOML parser is available in python stdlib for all versions of python flit intends to support

Yup, definitely. The vendored version will stay there as long as it's needed for any supported Python version.

PR #495, opened 2021-12-30
Import system tomli if vendor version not available

kitterma:
In Debian we try very hard to avoid vendoring, so I've patched flit 3.6.0 to use the system tomli.  Because of how our package archive is set up, we don't have the bootstrapping problem that is the reason you vendored it.  It would be nice if you would fall back to the system tomli so i didn't have to carry a patch for this.

### DIFF
--- a/flit_core/flit_core/config.py
+++ b/flit_core/flit_core/config.py
@@ -7,7 +7,10 @@ import os.path as osp
 from pathlib import Path
 import re
 
-from .vendor import tomli
+try:
+    from .vendor import tomli
+except ModuleNotFoundError:
+    import tomli
 from .versionno import normalise_version
 
 log = logging.getLogger(__name__)


pradyunsg (commented):
2 questions:

@kitterma why not patch the import on Debian’s end, if you’re already removing the vendored library?

@takluyver any strong feelings?

kitterma:
I did, but I prefer to minimize the amount of patching we have to do and would prefer to drop it if I can.  flit_core/flit_core/vendor/README mentions the possibility of de-vendoring.  The README doesn't mention the need to change the code, just the potential for cyclic dependencies.  This pull request actually makes it work the way the README suggests.

pradyunsg:
https://github.com/pypa/flit/blob/68ed8526b4293d4c8b9bc9bb41af07599a009db4/flit_core/flit_core/vendor/README#L6-L8

I think you and I might be reading this differently.

kitterma:
Agreed.  I was reading too quickly.  In any case I think there's no harm in this for users of the vendored module and it would make life easier for those that don't (which is why I picked to import from the vendored location first).  If you want, I can update the PR to update the README to remove the bit about changing the import.

pradyunsg:
If you’re modifying the package AKA patching it, it’s better to have any only-used-for-patched-code-paths code in your patch, instead of upstream.

That said, I’m not going to be blocking this PR from merging, if Thomas and Bernat are not opposed to this. I’m gonna now out now.

takluyver:
I knew bundling something would mean hearing from a Debian packager, but that's much faster than I expected. :wink: 

My leaning is to leave it as is, and require downstreams that want to un-bundle it to carry a small patch to import the system tomli. I guess I roughly figured you're 'patching' it anyway, even if the only modification is deleting files, so it's not much more onerous to add one line. Having a fallback always present might conceal some mistake I make bundling a future version - e.g. if tomli gains an external dependency, and I bundle the new version without noticing that. It's admittedly unlikely, but not impossible, so I'd prefer to have the simple import and make any failures more likely to show up.

This isn't a very strong preference, though - if the patch is an issue for you, I can live with the try/except fallback.

kitterma:
I think that makes sense.  It's not a significant issue to carry the patch.  I try to keep patches to a minimum, so I thought I'd ask.  I'll close this.  Thanks for the feedback.

takluyver:
Thanks Scott. :slightly_smiling_face:

PR #498, opened 2022-01-02
Lowercase distribution name in wheel filename and .dist-info folder name

takluyver:
In line with this correction in the wheel spec: https://github.com/pypa/packaging.python.org/pull/1032

The [spec for `.dist-info` directories](https://packaging.python.org/en/latest/specifications/recording-installed-packages/) already requires this, by referring to [PEP 503](https://www.python.org/dev/peps/pep-0503/#normalized-names).


pradyunsg (approved):

takluyver:
The spec change was merged, so I'm merging this as well. The change in Warehouse is still outstanding: https://github.com/pypa/warehouse/pull/10072

PR #506, opened 2022-01-07
Add option to fail on unknown keys.

### LINKED ISSUE #505: warn on pep 621 unknown keys ?
I just scratched my head for 30 minutes because of a typo. 

Do @takluyver think it could be ok to warn on extra keys ?

Carreau:
As a draft, config, and naming to discuss.

I'm tempted to make it strict by default, and have the option to ignore
if ever there are extra keys added.

Closes #505

### DIFF
--- a/flit_core/flit_core/config.py
+++ b/flit_core/flit_core/config.py
@@ -82,6 +82,8 @@ def prep_toml_config(d, path):
     """
     dtool = d.get('tool', {}).get('flit', {})
 
+    strict = dtool.get("config", {}).get("fail_on_unknown_pep621_keys", False)
+
     if 'project' in d:
         # Metadata in [project] table (PEP 621)
         if 'metadata' in dtool:
@@ -94,7 +96,7 @@ def prep_toml_config(d, path):
                 "[tool.flit.entrypoints]. Use [project.scripts],"
                 "[project.gui-scripts] or [project.entry-points] as replacements."
             )
-        loaded_cfg = read_pep621_metadata(d['project'], path)
+        loaded_cfg = read_pep621_metadata(d["project"], path, strict)
 
         module_tbl = dtool.get('module', {})
         if 'name' in module_tbl:
@@ -120,7 +122,12 @@ def prep_toml_config(d, path):
         )
 
     unknown_sections = set(dtool) - {
-        'metadata', 'module', 'scripts', 'entrypoints', 'sdist'
+        "metadata",
+        "module",
+        "scripts",
+        "entrypoints",
+        "sdist",
+        "config",
     }
     unknown_sections = [s for s in unknown_sections if not s.lower().startswith('x-')]
     if unknown_sections:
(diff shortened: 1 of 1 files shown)


takluyver:
I'd certainly be OK with having the strict option by default, and I think I'd go one step further and not make an option at all.

I just remembered we already have an environment variable `FLIT_ALLOW_INVALID` as an escape hatch for metadata that fails validation. That applies to a separate validation step which occurs after loading the config proper, but I don't know if that distinction is really important. Maybe that's close enough that we could reuse it for this, giving users a way round the problem if there is an unknown key, without needing another option? :shrug:

Carreau:
Ok, let me see. I want to test if the FLIT_ALLOW_INVALID pass the `pip install -e .` and `python -m build`. Those are my main concern. As things get more and more built in isolated envs, I'm still sort-of thinking that a value that is part of the pyproject.toml sort of make sens.

takluyver:
That's a good point. But I think a value in the pyproject.toml is perhaps not ideal, because that's controlled by the maintainers of the package, not by the people who might be trying to build it. Of course, the people downstream could patch `pyproject.toml` to use this option... but then they could also patch it to remove the extra keys.

Carreau:
I agree, I just don't see a reason to set/unset it outside of `pyproject.toml`, cause if there are extra keys then the maintainers should be aware of them.

(I actually don't see a reason to unset failing on non-existing key at all unless there are new keys and an old versions of flit...).

pradyunsg:
> I want to test if the FLIT_ALLOW_INVALID pass the `pip install -e .` and `python -m build`.

I can confirm that it will work. :)

I use the same mechanism in another build backend: https://github.com/pradyunsg/sphinx-theme-builder/blob/97bd61c3a441341cd33c5af11c3264226c297b9e/src/sphinx_theme_builder/_internal/nodejs.py#L99

takluyver:
Thanks @pradyunsg . :slightly_smiling_face: 

I'm also OK with just erroring on unknown keys with no escape hatch - I think the PEP 621 support is new enough that we can probably get away with this, and we can always add an escape hatch if it turns out to be an issue. If we do want an escape hatch, the environment variable is my preference.

takluyver:
It appears that at least one package has made it to PyPI with a misspelled key - `pandas_dataframe_convert==0.2` has `dependancies` instead of `dependencies`.

This is probably a prompt to get this fixed quickly. But maybe also that we do need some kind of escape hatch to let such packages be built. :confused:

Carreau:
Well at the same time if it’s misspelled it’s build will be incorrect.

takluyver:
Technically incorrect but still largely working. Given what it is, I'd guess a lot of users have `pandas` installed anyway.

It was only up for a matter of hours before a new release corrected the issue, so in this case we'll probably get away with breaking the older release. But it does show that real people are hitting the issue and managing to upload packages to PyPI.

Carreau:
> Technically incorrect

The best kind of incorrect.

takluyver:
My current thinking is that we'll make unknown `[project]` keys an error for Flit 4.0. I encourage all projects to specify an upper bound (currently `<4`) on the version of `flit_core` they build with, so that we can make changes like this without breaking existing packages.

PR #510, opened 2022-01-15
External data directory

### LINKED ISSUE #358: Support for data files installation
It would be handy to have some analogue of setuptools's `data_files` entry to be
able to configure the installation of data files (or did I miss it?) .

Thanks in advance!

takluyver:
Corresponding to setuptools' `data_files`, but less flexible: you just get one directory, which will be mapped to wherever the installer decides to put the 'data' component of the wheel (typically the prefix path).

Closes #358.

### DIFF
--- a/doc/pyproject_toml.rst
+++ b/doc/pyproject_toml.rst
@@ -428,6 +428,8 @@ Exclusions have priority over inclusions.
 External data section
 ---------------------
 
+.. versionadded:: 3.7
+
 Data files which your code will use should go inside the Python package folder.
 Flit will package these with no special configuration.
 
@@ -450,6 +452,11 @@ depends on how your package is installed and how those systems are configured.
 For instance, installing in a virtualenv usually doesn't affect anything outside
 that environment. Don't rely on these files being picked up unless you have
 close control of how the package will be installed.
+
+If you install a package with ``flit install --symlink``, a symlink is made
+for each file in the external data directory. Otherwise (including development
+installs with ``pip install -e``), these files are copied to their destination,
+so changes here won't take effect until you reinstall the package.
 
 .. note::
 
--- a/doc/pyproject_toml.rst
+++ b/doc/pyproject_toml.rst
@@ -417,4 +417,35 @@ These paths:
 
 Exclusions have priority over inclusions.
 
+External data section
+---------------------
+
+Data files which your code will use should go inside the Python package folder.
+Flit will package these with no special configuration.
+
+However, sometimes it's useful to package external files for system integration,
+such as man pages or files defining a Jupyter extension. To do this, arrange
+the files within a directory such as ``data``, next to your ``pyproject.toml``
+file, and add a section like this:
+
+.. code-block:: toml
+
+    [tool.flit.external-data]
+    directory = "data"
+
+Paths within this directory are typically installed to corresponding paths under
+a prefix (such as a virtualenv directory). E.g. you might save a man page for a
+script as ``(data)/share/man/man1/foo.1``.
+
+Whether these files are detected by the systems they're meant to integrate with
+depends on how your package is installed and how those systems are configured.
+For instance, installing in a virtualenv usually doesn't affect anything outside
+that environment. Don't rely on these files being picked up unless you have
+close control of how the package will be installed.
+
+.. note::
+
+   For users coming from setuptools: external data corresponds to setuptools'
+   ``data_files`` parameter, although setuptools offers more flexibility.
+
 .. _environment marker: https://www.python.org/dev/peps/pep-0508/#environment-markers
--- a/flit_core/flit_core/common.py
+++ b/flit_core/flit_core/common.py
@@ -421,8 +421,8 @@ def dist_info_name(distribution, version):
 def walk_data_dir(data_directory):
     """Iterate over the files in the given data directory.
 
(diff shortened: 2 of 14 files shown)


doc/pyproject_toml.rst
```diff
@@ -417,4 +417,35 @@ These paths:
...
 Exclusions have priority over inclusions.
 
+External data section
+---------------------
+
+Data files which your code will use should go inside the Python package folder.
+Flit will package these with no special configuration.
+
+However, sometimes it's useful to package external files for system integration,
+such as man pages or files defining a Jupyter extension. To do this, arrange
```
bollwyvl, line 427: perhaps

```rst
which rely on well-known paths such as ``man`` pages or defining a Jupyter `extension <https://jupyter-server.readthedocs.io/en/latest/developers/extensions.html>`_ or `kernel <https://jupyter-client.readthedocs.io/en/stable/kernels.html#kernel-specs>`_. To do this, arrange
```

> :blush: sorry, I'm sure there's a better way to do this in `.rst`...

doc/pyproject_toml.rst
```diff
@@ -417,4 +417,35 @@ These paths:
 
 Exclusions have priority over inclusions.
 
+External data section
+---------------------
+
```
bollwyvl, line 422: Maybe with a...

```rst
.. versionadded:: ???
```

doc/pyproject_toml.rst
```diff
@@ -417,4 +417,35 @@ These paths:
...
+depends on how your package is installed and how those systems are configured.
+For instance, installing in a virtualenv usually doesn't affect anything outside
+that environment. Don't rely on these files being picked up unless you have
+close control of how the package will be installed.
+
+.. note::
+
+   For users coming from setuptools: external data corresponds to setuptools'
+   ``data_files`` parameter, although setuptools offers more flexibility.
+
```
bollwyvl, line 450: Perhaps a note/warning on how/if-at-all this works with editable installs, and how one might go about handling developing these?

takluyver: Thanks, I'm adding a note. They'll get installed, but with a `pip install -e` development install, they're just copied, so you have to reinstall if anything changes (I don't think PEP 660 really lets us do anything else). If you use `flit install --symlink`, individual files are symlinked, so you only have to reinstall when adding/removing them.

doc/pyproject_toml.rst
takluyver: I'm torn on this, but for now I think I'll leave the links out. Jupyter has quite a few different extension points, and I don't want this to grow into a list of ways to extend Jupyter (not that such a list shouldn't exist, but it should be in Jupyter's own docs ;-).

takluyver:
I'll leave this another day or two for any other maintainers (or anyone else!) to look at, otherwise I think it's ready.

doc/pyproject_toml.rst
bollwyvl: Yep, sounds good!

blink1073 (approved):
Fantastic, thanks @takluyver!

flit_core/flit_core/common.py
```diff
@@ -416,3 +416,20 @@ def normalize_dist_name(name: str, version: str) -> str:
 def dist_info_name(distribution, version):
     """Get the correct name of the .dist-info folder"""
     return normalize_dist_name(distribution, version) + '.dist-info'
+
+
+def walk_data_dir(data_directory):
+    """Iterate over the files in the given data directory.
+
+    Yields absolute paths - caller may want to make them relative.
```
pradyunsg, line 424: If the argument passed in is `./foo`, will it expand that to `/Users/pradyunsg/Developer/magic-project/foo`? If not, I think we might want to rephrase this to say paths are always prefixed with `data_directory` or soemthing like that.

flit_core/flit_core/config.py
```diff
@@ -142,6 +142,27 @@ def prep_toml_config(d, path):
             dtool['sdist'].get('exclude', []), 'exclude'
         )
 
+    data_dir = dtool.get('external-data', {}).get('directory', None)
+    if data_dir is not None:
+        toml_key = "tool.flit.external-data.directory"
+        if not isinstance(data_dir, str):
+            raise ConfigError(f"{toml_key} must be a string")
```
pradyunsg, line 149: Do you want to add tests for these error branches, that codecov is complaining aren't covered?

pradyunsg (commented):
This is a review from a sleepy-me. I'll try to take a proper look over the weekend.

flit_core/flit_core/common.py
takluyver: Yup, good point.

flit_core/flit_core/config.py
takluyver: It feels too much like testing that the code I've written is the code that's there, which always feels a bit pointless. If someone wants to add some test cases and push coverage up a bit, I'm happy enough to merge it, but I don't feel like it's worth spending time on.

If someone finds a case where it's currently doing the wrong thing, I'm happier to add a test for that as we fix it.

maartenbreddels:
Just want to say, perfect. Simple solution that does exactly what I want, no complicated mapping/rename etc, just plain and simple 👍

PR #511, opened 2022-01-16
Add bootstrap script/wheel module for flit_core

jameshilliard:
Improved version of #481 based on comments [here](https://github.com/pypa/build/issues/430#issuecomment-1008874876).

Details from help:
```
$ python -m flit_core.wheel --help
usage: wheel.py [-h] [--srcdir SRCDIR] [--outdir OUTDIR]

options:
  -h, --help            show this help message and exit
  --srcdir SRCDIR, -s SRCDIR
                        source directory (defaults to current directory)
  --outdir OUTDIR, -o OUTDIR
                        output directory (defaults to /Users/jameshilliard/flit/flit_core/dist)
```
```
$ python bootstrap.py --help
usage: bootstrap.py [-h] {build,install} ...

positional arguments:
  {build,install}

options:
  -h, --help       show this help message and exit
```
```
$ python bootstrap.py build --help
usage: bootstrap.py build [-h] [--outdir OUTDIR]

options:
  -h, --help            show this help message and exit
  --outdir OUTDIR, -o OUTDIR
                        output directory (defaults to /Users/jameshilliard/flit/flit_core/dist)
```
```
$ python bootstrap.py install --help
usage: bootstrap.py install [-h] [--wheeldir WHEELDIR] [--installdir INSTALLDIR]

options:
  -h, --help            show this help message and exit
  --wheeldir WHEELDIR, -w WHEELDIR
                        wheel dist directory (defaults to /Users/jameshilliard/flit/flit_core/dist)
  --installdir INSTALLDIR, -i INSTALLDIR
                        installdir directory (defaults to /Users/jameshilliard/flit/venv/lib/python3.10/site-packages
```

### DIFF
--- a/flit_core/bootstrap.py
+++ b/flit_core/bootstrap.py
@@ -3,85 +3,39 @@
 Normally, you would install flit_core with pip like any other Python package.
 This script is meant to help with 'bootstrapping' other packaging
 systems, where you may need flit_core to build other packaging tools.
-
-Pass a path to the site-packages directory or equivalent where the package
-should be placed. If omitted, this defaults to the site-packages directory
-of the Python running the script.
 """
 import argparse
 import sys
 import sysconfig
 from pathlib import Path
-from tempfile import TemporaryDirectory
 from zipfile import ZipFile
 
-from flit_core.wheel import add_wheel_arguments, build_flit_wheel
-
-srcdir = Path(__file__).parent.resolve()
-
 def extract_wheel(whl_path, dest):
-    print("Installing to", dest.resolve())
+    print("Installing to", dest)
     with ZipFile(whl_path) as zf:
         zf.extractall(dest)
 
-def add_install_arguments(parser):
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
     parser.add_argument(
-        '--wheeldir',
-        '-w',
-        type=str,
-        help=f'wheel dist directory (defaults to {srcdir.joinpath("dist").resolve()})',
+        'wheel',
+        type=Path,
+        help=f'flit_core wheel to install (.whl file)',
     )
-    purelib = Path(sysconfig.get_path('purelib'))
+    purelib = Path(sysconfig.get_path('purelib')).resolve()
     parser.add_argument(
         '--installdir',
         '-i',
         type=Path,
         default=purelib,
-        help=f'installdir directory (defaults to {purelib.resolve()}',
+        help=f'installdir directory (defaults to {purelib})',
     )
-    return parser
-
-def get_dist_wheel(wheeldir):
-    wheel_path = Path(wheeldir) if wheeldir is not None else srcdir.joinpath('dist')
-    wheel_glob = wheel_path.glob('flit_core-*.whl')
-    return next(wheel_glob, None)
-
-def build(args):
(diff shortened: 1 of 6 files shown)


flit_core/bootstrap.py
```diff
@@ -0,0 +1,87 @@
...
+        zf.extractall(dest)
+
+def add_install_arguments(parser):
+    parser.add_argument(
+        '--wheeldir',
+        '-w',
+        type=str,
+        help=f'wheel dist directory (defaults to {srcdir.joinpath("dist").resolve()})',
+    )
+    site_packages = Path(sysconfig.get_path('purelib'))
```
FFY00, line 34: nitpick: Please call this `purelib`, site packages is a different concept.

FFY00:
Ah, I think I missed something, shouldn't this be installed? If so, this should live under the `flit_core` module, and accessible via `python -m flit_core.bootstrap`.

I will look at this in more detail later.

jameshilliard:
> Ah, I think I missed something, shouldn't this be installed?

Only the wheel builder is installed since the `bootstrap.py` installer is only designed to install `flit_core` itself, it's expected that [`installer`](https://github.com/pradyunsg/installer) be used to install wheels built using `python -m flit_core.wheel`. This should work for bootstrapping since [`installer`](https://github.com/pradyunsg/installer) only depends on `flit_core`.

flit_core/bootstrap.py
jameshilliard: changed to `purelib`

takluyver:
Thanks for this. In the interests of keeping the extra script as simple as possible, and minimising coupling between the script and the internals of flit_core, I'd like to have the script only do installation, and use `python -m flit_core.wheel` directly to build a wheel. From what we've discussed this is the same command you'll use to build wheels for a few subsequent packages (e.g. `installer`, `tomli`), so I don't think it should be a problem to use it for `flit_core` itself as well.

I know what I want this to look like, so I hope you won't mind if I just make the changes directly rather than asking you to do them.

takluyver:
OK, I hope this works for you.

I've got rid of the globbing inside the script - if you know there's only one suitable wheel in a directory, you can always use `dist/flit_core-*.whl` in a script calling it. And I've got rid of the fallback to building a wheel, because I want this to do just one simple thing, not have different code paths. You've mentioned in the past that you separate build & install steps, so I assume you don't need one script combining them (and if anyone does, it should be easy to write that themselves).

jameshilliard:
> In the interests of keeping the extra script as simple as possible, and minimising coupling between the script and the internals of flit_core, I'd like to have the script only do installation, and use `python -m flit_core.wheel` directly to build a wheel.

Yeah, I don't think that should be an issue, I had just attempted to preserve the combo build+install option from #481 but for our use case where we do build+install stage separation a combo build+install is not actually needed.

> I know what I want this to look like, so I hope you won't mind if I just make the changes directly rather than asking you to do them.

> OK, I hope this works for you.

Yeah, I think this should work for us.

takluyver:
Great, thanks. I've updated the bootstrapping doc. I'll give it another day or two in case other maintainers want to look at it, but otherwise I think this can go in.

blink1073 (approved):
LGTM!

Issue #513, opened 2022-01-20
`flit build` packages different files into the sdist than `python -m build`


BubuOT:
I have a package that is build using `flit-core` (actually via [`flit-scm`](https://pypi.org/project/flit-scm/)). When running `flit build` it follows the [documentation](https://flit.readthedocs.io/en/latest/pyproject_toml.html#sdist-section) in that it packages all git tracked files in the repository. When doing the build via `python -m build` it instead gathers only the files in our modules subdir though I don't really ever specified that? (that i.e. doesn't include the tests, which I'd like to ship in the sdist as well). 

The `[tool.flit.sdist]` `include` and `exclude` directives can be used to work around that but it seems odd that I'd need a different set of directives between different build frontends.

Is this behaviour to be expected for some reason or is this a bug?

takluyver:
It's expected for now, though I'd like to figure out a way out of it.

Flit was around before PEP 517 and the neat separation of frontends and backends. Back then, I saw sdists as just a way to publish a snapshot of the source code on PyPI, so only a package's maintainer would ever need to make them. In that context, relying on git (or another VCS) to decide which files to include made sense.

Other people took a different view, and making an sdist was standardised as a backend hook, so a frontend can ask the backend to make an sdist in any context. It didn't make sense for this backend operation to require either VCS tools (the `git` command) or VCS metadata (the `.git` folder). The compromise I chose was for the backend to make a 'minimal' sdist, with just the files needed to build a wheel, but for `flit build` to carry on using the VCS and making a larger sdist, like it always had, for publishing on PyPI.

That probably needs to change at some point, because people obviously expect that an sdist is an sdist. But it mostly works for what I do, so it's not a top priority for me. The VCS integration has been a source of complexity and bugs, so I'd be tempted to get rid of it and just let people use `[tool.flit.sdist]` to control what's included.

BubuOT:
I see, thanks for the explanation. Do you think it makes sense to add a note about these behaviour differences to https://flit.readthedocs.io/en/latest/pyproject_toml.html#sdist-section?

takluyver:
Yes, I think a note there would be helpful.

takluyver:
Closed by #514

PR #514, opened 2022-01-20
docs: mention flit-core sdist behaviour

BubuOT:
Ref: https://github.com/pypa/flit/issues/513#issuecomment-1017407065

### DIFF
--- a/doc/pyproject_toml.rst
+++ b/doc/pyproject_toml.rst
@@ -417,4 +417,13 @@ These paths:
 
 Exclusions have priority over inclusions.
 
+.. note::
+
+   If you are not using :ref:`build_cmd` but  ``flit_core`` via another build
+   frontend Flit doesn't doesn't check the VCS for files to include but instead
+   builds a 'minimal' sdist (which includes the files necessary to build a wheel).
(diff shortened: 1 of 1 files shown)


doc/pyproject_toml.rst
```diff
@@ -417,4 +417,13 @@ These paths:
 
 Exclusions have priority over inclusions.
 
+.. note::
+
+   If you are not using :ref:`build_cmd` but  ``flit_core`` via another build
+   frontend Flit doesn't doesn't check the VCS for files to include but instead
```
takluyver, line 423:
```suggestion
   frontend, Flit doesn't doesn't check the VCS for files to include but instead
```

doc/pyproject_toml.rst
```diff
@@ -417,4 +417,13 @@ These paths:
...
+   builds a 'minimal' sdist (which includes the files necessary to build a wheel).
+   You'll have to adaopt your inclusion/exclusion rules to achieve the same result
```
takluyver, line 425:
```suggestion
   You'll have to adapt your inclusion/exclusion rules to achieve the same result
```

takluyver:
Thanks!

PR #516, opened 2022-01-21
flit_core: refactor path handling to use pathlib

jameshilliard:
This should make path handling easier/cleaner compared with using `os.path`.


flit_core/flit_core/common.py
```diff
@@ -66,7 +66,7 @@ def __init__(self, name, directory=Path()):
     @property
     def file(self):
         if self.is_package:
-            return self.path / '__init__.py'
+            return self.path.joinpath('__init__.py')
```
pradyunsg, line 69: Could you please remove these `/` -> joinpath changes? I don't think they're particularly beneficial, and seem to be the vast majority of this PR.

flit_core/flit_core/common.py
```diff
@@ -21,11 +21,11 @@ def __init__(self, name, directory=Path()):
         self.name = name
 
         # It must exist either as a .py file or a directory, but not both
-        name_as_path = name.replace('.', os.sep)
-        pkg_dir = directory / name_as_path
-        py_file = directory / (name_as_path+'.py')
-        src_pkg_dir = directory / 'src' / name_as_path
-        src_py_file = directory / 'src' / (name_as_path+'.py')
+        name_as_path = Path(name.replace('.', os.sep))
```
pradyunsg, line 24: This is unnecessary.

flit_core/flit_core/buildapi.py
```diff
@@ -48,36 +45,36 @@ def prepare_metadata_for_build_wheel(metadata_directory, config_settings=None):
...
 def build_editable(wheel_directory, config_settings=None, metadata_directory=None):
     """Builds an "editable" wheel, places it in wheel_directory"""
-    info = make_wheel_in(pyproj_toml, Path(wheel_directory), editable=True)
+    info = make_wheel_in(pyproj_toml, wheel_directory, editable=True)
     return info.file.name
 
 def build_sdist(sdist_directory, config_settings=None):
     """Builds an sdist, places it in sdist_directory"""
-    path = SdistBuilder.from_ini_path(pyproj_toml).build(Path(sdist_directory))
+    path = SdistBuilder.from_ini_path(pyproj_toml).build(sdist_directory)
```
pradyunsg, line 79: I'd much rather that the various internals use `Path` objects, which are created here. That's much easier to reason about than thinking about the `str` objects coming in via this API.

flit_core/flit_core/config.py
```diff
@@ -201,18 +231,18 @@ def _check_glob_patterns(pats, clude):
                 .format(clude, p)
             )
 
-        normp = osp.normpath(p)
+        normp = posix_normpath(PurePosixPath(p))
```
pradyunsg, line 234: This looks like a behaviour change on Windows -- since it's switching from Windows paths to POSIX paths.

pradyunsg (changes requested):

pradyunsg:
I'm not sure that it is particularly useful to make such a migration, but I'll leave the merits of thinking about those to others.

flit_core/flit_core/config.py
jameshilliard: The path handling seemed buggy/inconsistent here since os.path.normpath is filesystem/os specific and these validations I think need to be consistent across platforms.

flit_core/flit_core/buildapi.py
jameshilliard: Fixed, forgot to revert that from a test I was doing.

flit_core/flit_core/common.py
jameshilliard: removed

flit_core/flit_core/common.py
jameshilliard: I changed that part to ensure `Path` types were always being used in those locations properly, the implementation between `/` and joinpath appears to be [slightly different](https://github.com/python/cpython/blob/v3.10.2/Lib/pathlib.py#L843-L861). I find joinpath to generally be preferable as it's more explicit but that's somewhat subjective.

pradyunsg: Well, it adds a lot of noise to this PR. I'm unlikely to re-review this PR, if this style of change isn't dropped. :)

jameshilliard: Ok, I'll back those parts out.

jameshilliard: > Could you please remove these `/` -> joinpath changes?

I think I have all of these that weren't needed backed out now.

flit_core/flit_core/config.py
jameshilliard: The tests already expect these paths to always be posix style at the moment, as this path is not an absolute path on windows:
https://github.com/pypa/flit/blob/6982faee6f69e21327bdc8cf245fe99d57ea9fc3/flit_core/flit_core/tests/test_config.py#L109

PR #531, opened 2022-02-22
Fix relative imports in __init__.py when exec-ing to get __version__

### LINKED ISSUE #530: ModuleNotFoundError: No module named 'flit_core.dummy'
I'm building my project with
```toml
[build-system]
requires = ["flit_core >=3.2,<4"]
```
and since earlier today I'm getting the error message
```
      [...]
      ModuleNotFoundError: No module named 'flit_core.dummy'
      [end of output]
```
when running `pip install .` on the repo. Perhaps I messed up something in the package, but the error message points towards flit. Any idea what's going on?

takluyver:
It appears that the module needs to be in `sys.modules` for relative imports to work. The deprecated `.load_module()` API did this for you, but `.exec_module()` doesn't.

`importlib` has a [bunch of extra complexity](https://github.com/python/cpython/blob/77f31a91d55df2df79ac767690738081f27ad476/Lib/importlib/_bootstrap.py#L678-L705) in the code path implementing `.load_module()`, but as far as I can tell that's all about thread safety, and Flit should never be running in a multithreaded context, so I've ignored that.

Fixes #530


pradyunsg (approved):

pradyunsg:
@takluyver Let's get this landed and released? We've got a decent number of users flagging this as an issue and needing to work around it.

PR #537, opened 2022-03-22
Output license.text to license field if supplied in metadata

ccorbacho:
Until PEP 639 is finalized, many projects rely on the output of License
in PKG-INFO being an SPDX string, if it is using a standard license,
and third party tooling can process that.

Rather than silently ignoring license.text, use that to populate the
License field if it is provided, and leave it to downstream packages as
to whether they are supplying license.file or license.text


AA-Turner:

A

takluyver:
True, thank-you.

PR #546, opened 2022-04-26
Add "install-reqs" command to CLI

dciborow:
This extends the CLI, and provides a command wrapper for an additional public function, `install_requirements()`. This will skip installing the user's code, but will still install all required dependencies. This is particularly useful for preparing building images or containers by enabling users to pre-install dependencies, without installing the source code directly.

This enables installation of dependencies without the need for separate requirements.txt files

`flit install-reqs`

The following command can be used instead of `pip install -r requirements.test.txt` or `pip install .[test]`

`flit install-reqs --deps test`

Open Issues
The following issues are present, because they are required for normal package installation, but are unnecessarily when installing the dependencies.
- User is required to have "README.md"
- User is required to have package directory

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -77,13 +77,11 @@ To install a package locally for development, run::
 
 Flit packages a single importable module or package at a time, using the import
 name as the name on PyPI. All subpackages and data files within a package are
-included automatically.
-
-To install only the package's requirements if preparing a container or image, run::
+included automatically. To only install only the package's requirements if preparing a container or image, run::
 
     flit install-reqs [--deps 'all']
 
-This is the equivlant of running 'pip install -r requirements.txt'.
+This is the similiar to the previously common technique of running 'pip install -r requirements.txt'.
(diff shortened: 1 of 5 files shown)


dciborow:
@pradyunsg , I am coming out of no where with this suggested improvement. 

I have been working with @brettcannon on a template for Python Projects using pyproject.toml. When converting, I removed my `requirements.txt` and `requirements.test.txt`, which would previously be parsed in my `setup.py`. This leaves me a bit stuck on how to install just my dependencies. Before I went and created some crazy TOML parser, i discovered all the needed code was already located in a public method within 'flit', but had not been exposed.

dciborow:
@takluyver do you have an feedback for this requested feature?

svartkanin:
Is anything blocking this from getting included?

takluyver:
I've gone with `--only-deps` - thanks!

dciborow:
changing the title so it accurate reflects the implementation.

PR #550, opened 2022-05-13
Add support for recursive `**` includes and excludes

layday:
Signed-off-by: layday <layday@protonmail.com>


takluyver:
Thanks, this looks good to me!

Issue #149, opened 2017-11-16
Dash in dist-name


asvetlov:
I'm working on https://github.com/wikibusiness/aiohttp-remotes library

The library has a dash in the name but top level package is obviously named with underscore.
`flit` supports `dist-name` option: `dist-name="aiohttp-remotes"`

In this case `flit build` produces correct wheel and tarball names but `pip install` is screwed.
It actually installs the distribution (I could do import it etc.) but `pip freeze` don't see `aiohttp-remotes` installed in virtual environment, `pip uninstall aiohttp-remotes` fails -- no such library was installed from `pip` perspective.

As partial solution I've dropped `dist-name` directive from `pyproject.toml`.
What is recommended way for supporting dashes instead of underscores in package names?

takluyver:
It looks like the `.dist-info` folder needs the name to be normalised to use underscores. You're not doing anything wrong, it's just a bug in flit.

I think it also doesn't matter whether you use `-` or `_` on PyPI - they'll both work for installing the package anyway. But we should make sure it works properly.

takluyver:
#150 should clear it up.

asvetlov:
Recently linux package maintainers asked me to use dashes in library names (`aiohttp-sesssion` for example).
IIRC some linux distros use PyPI tarballs for automatic repackaging.
Is it possible to keep `-` in tarball and replace it with `_` for wheel -- as `setuptools` already does?

P.S. 
I'm trying to use `flit` in my new pure python projects and very happy with it.
Much less dances with deployment comparing to `setup.py`.
Very thank you and please keep the project going.
Supporting C Extensions is another pain, maybe I should try enscons

takluyver:
Once #150 is merged and released, you should be able to go back to what you were doing (`dist-name = "aiohttp-remotes"`), and it will hopefully work as you expect.

I'm puzzled why Linux distro maintainers should care, though. We have several Jupyter/IPython packages with underscores in the name (e.g. [jupyter_core](https://pypi.python.org/pypi/jupyter_core)), and I don't recall anyone asking us to do it differently. It makes more sense to me to do it that way, because then the distribution name matches the import name.

takluyver:
PS Thanks, I'm glad you're enjoying using Flit!

asvetlov:
I don't know.
Perhaps it depends on distribution tooling.
Anyway the change was easy for me and didn't break anything -- so why not?

takluyver:
[...]

PR #566, opened 2022-07-11
Normalize module name default value

joukewitteveen:
The package name is used as the default module name. Technically, the
module name must be normalized (no hyphens), so for convenience set the
default module name to the normalized package name, with hyphens replaced
by underscores.

---
Normalization is mentioned in [PEP426](https://peps.python.org/pep-0426/#name) and came up before in #149. ~The change proposed here was not tested (it is proposed to gauge interest and sentiment). Additionally, the documentation is not touched, but I feel normalization may be left implicit since~ a package name with hyphens isn't a valid import name to begin with.

_edit:_ To clarify, with this change the default module name for the `aiohttp-remotes` project becomes `aiohttp_remotes`.

### DIFF
--- a/doc/pyproject_toml.rst
+++ b/doc/pyproject_toml.rst
@@ -69,8 +69,9 @@ The allowed fields are:
 
 name
   The name your package will have on PyPI. This field is required. For Flit,
-  this also points to your package as an import name by default (see
-  :ref:`pyproject_module` if that needs to be different).
+  this name, with any hyphens replaced by underscores, is also the default value
+  of the import name (see :ref:`pyproject_module` if that needs to be
+  different).
 version
   Version number as a string. If you want Flit to get this from a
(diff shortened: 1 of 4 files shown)


joukewitteveen:
I tested the proposed change and it seems to work well! I also updated the documentation anyway, including a little coverage of #260 while I was at it.

takluyver:
Part of my thinking with using the same field for both by default was to encourage people to use a single name for the import and the install - even as far as encouraging `test_package` over `test-package`. But I can see the benefit of allowing a broader set of names to 'just work', especially since pip lets you use `-` and `_` interchangeably in package names anyway.

So I'm inclined to accept this. :slightly_smiling_face: I'll give other maintainers a few days to chime in, otherwise I think it's ready to merge.

takluyver:
Thanks!

mohdzahidabdullah:
python --version

Issue #568, opened 2022-07-16
flit_core does not filter out Python bytecode when not operating on vcs


nanonyme:
[When flit_](https://github.com/pypa/flit/blob/ec6e6b7c8994d5e79d0fd951d9e4c26a7c5d9977/flit_core/flit_core/sdist.py#L109-L118) is allowing does not invoke https://github.com/pypa/flit/blob/ec6e6b7c8994d5e79d0fd951d9e4c26a7c5d9977/flit/sdist.py#L98-L101 on output files which results in Python bytecode ending up in package. This is almost certainly undesired behaviour as there is no portability with Python bytecode across other Python versions.

takluyver:
Can you describe a specific scenario where this is going wrong? The filtering works in somewhat different contexts, and I often live with little bits of duplication rather than tying modules pieces more closely together.

layday:
When building with flit-core, `__pycache__` folders are only excluded from the "module" folder, so if you were to add e.g. your tests folders in the sdist, `tests/__pycache__` won't be filtered out. IIUC, the flit builder excludes these from everywhere in the sdist.

takluyver:
Aha, I see. So the `_include` function is fine, because `Module.iter_files()` is only going through the files in the importable module - other files you configure to include in the sdist are added by `SdistBuilder.apply_includes_excludes()`.

Maybe the simplest way to do it is to always add `**/__pycache__` and `**.pyc` to the exclude patterns? Excludes override inclues, so there would be no way to configure including them, but I can't think of a reason you'd need that (and Flit doesn't aim to cover every possible scenario anyway).

nanonyme:
@takluyver do we really need also the directory? Does flit package empty directories?

PR #569, opened 2022-07-16
Unify pyc file calculation between two sdist implementations

nanonyme:
Fixes https://github.com/pypa/flit/issues/568


layday:
This doesn't fix it for non-package folders of sdists built with flit-core - you'll want to take the check out from both `flit_core/common.py` and `flit/sdist.py` and register both patterns in flit-core's `SdistBuilder` as default excludes.

takluyver:
Thanks, but as @layday noted, this doesn't actually fix it. I'll close this and continue discussion on #568, or any new PRs that come from that.

Issue #570, opened 2022-07-17
Feature requests for bootstrap install


nanonyme:
https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/blob/master/elements/components/python3-flit-core.bst#L14-17 is how we use this right now. (this is real meta distro usage of flit in real bootstrap scenario starting from plain Python). Any chance we could have following:
1. Prefix given as parameter (this is quite common that distros install under a prefix before creating a package)
2. Python bytecode compilation (optional?). Parameter could be eg -o given multiple times perhaps defaulting to current if none is given

It would allow getting rid of the boilerplate around calling the installer.

nanonyme:
Tested briefly and it looks like the platlib vs purelib difference doesn't matter for us. Prefix does. Our root is immutable during build except for very specific locations.

nanonyme:
Hm, also installer needs to create the destination directory with os.makedirs(path, exist_ok=True) if prefix is given. It most likely does not.

nanonyme:
Please do not call compileall with concurrency, it may break reproducible buils.

takluyver:
I'd really like to keep the bootstrap script as simple as possible, and leave e.g. byte compilation up to the caller to invoke separately.

In particular, there are [quite a few options](https://docs.python.org/3/library/compileall.html) for producing the .pyc files, and if the bootstrap install script calls this for you, it probably needs to re-expose quite a few of those, which builds further complexity.

To be honest, your four line snippet looks like pretty much what I'm expecting distros to do. It's a shame that there isn't a shorter way to get the purelib/platlib directory at the command line, but it's still a one liner. We could probably do the `mkdir -p` step inside the bootstrap script, which would save you one line.

I'll copy the snippet here for reference, in case it changes in the future:

```shell
PLATLIB=$(python -c "print(__import__('sysconfig').get_path('platlib'))")
mkdir -p %{install-root}$PLATLIB
python bootstrap_install.py dist/*.whl --installdir %{install-root}$PLATLIB
python3 -m compileall -o 0 -o 1 -o 2 "%{install-root}$PLATLIB"
```

nanonyme:
I think this is fine. So essentially now the minimal amount of boilerplate is
```
cd flit_core
python3 -mflit_core.whel
python3 bootstrap_install.py dist/*.whl --install-root ${install-root}
python3 -m compileall -o 0 -o 1 --o 2 ${install-root}
```

PR #571, opened 2022-07-17
Support optional installation root

nanonyme:
Part of https://github.com/pypa/flit/issues/570

### DIFF
--- a/flit_core/bootstrap_install.py
+++ b/flit_core/bootstrap_install.py
@@ -37,12 +37,21 @@ if __name__ == "__main__":
         default=purelib,
         help=f'installdir directory (defaults to {purelib})',
     )
+    parser.add_argument(
+        '--install-root',
+        type=Path,
+        default=None,
+        help='if given, installdir is considered to be under this'
+    )
 
     args = parser.parse_args()
 
     if not args.wheel.name.startswith('flit_core-'):
         sys.exit("Use this script only for flit_core wheels")
-    if not args.installdir.is_dir():
-        sys.exit(f"{args.installdir} is not a directory")
+    if args.install_root:
+        installdir = args.install_root / args.installdir.relative_to("/")
+    else:
+        installdir = args.installdir
 
-    extract_wheel(args.wheel, args.installdir)
+    installdir.mkdir(parents=True, exist_ok=True)
+    extract_wheel(args.wheel, installdir)


nanonyme:
@takluyver any chance getting opinions/reviews about this and the sibling PR https://github.com/pypa/flit/pull/572?

takluyver:
This one I might accept, but not with the name 'prefix', because `sys.prefix` and pip's `--prefix` argument both mean something different. I think the semantics of this match pip's `--root` option or [installer's](https://installer.readthedocs.io/en/stable/cli/installer/) `--destdir`.

Is this still useful to you without #572? Perhaps we could print the target path to stdout so you could do something like:

```shell
INSTALLED_DIR=$(python bootstrap_install.py ...)
python -m compileall $INSTALLED_DIR
```

nanonyme:
I will need to experiment a bit. I don't like the stdout-as-api approach. Perhaps I can just tell it to operate on install-root recursively, it should not contain content that does not need to be have bytecode compiled.

nanonyme:
Yes, this would be useful standalone. it's very minor extra work compileall has to do to traverse entire install-root.

flit_core/bootstrap_install.py
```diff
@@ -37,12 +37,23 @@ def extract_wheel(whl_path, dest):
...
 
     if not args.wheel.name.startswith('flit_core-'):
         sys.exit("Use this script only for flit_core wheels")
-    if not args.installdir.is_dir():
-        sys.exit(f"{args.installdir} is not a directory")
-
-    extract_wheel(args.wheel, args.installdir)
+    if args.install_root:
+        installdir = args.install_root / args.installdir.relative_to("/")
+        installdir.mkdir(parents=True, exist_ok=True)
```
takluyver, line 53: Let's do the mkdir in either case, for consistency. I don't think there was a particular reason why I was checking for an existing directory - maybe I just wanted to ensure it failed with a clear error if the expectations weren't met.

nanonyme: We have to, the point with install-root is that install-root *does* exist but installdir *does not*. This is essentially a vanilla destination that is captured and packaged further.

nanonyme: I see. Well, if I can remove the check, then that would maybe work too.

takluyver:
Thanks!

PR #572, opened 2022-07-17
Optionally build bytecode for installed code

nanonyme:
Part of https://github.com/pypa/flit/issues/570, depends on https://github.com/pypa/flit/pull/571


takluyver:
As described on #570, I'm strongly inclined to decline this one, and ask callers to run `python -m compileall` themselves (as you already do). Sorry about that.

Issue #544, opened 2022-04-21
Use tomllib on Python 3.11+


gitpushdashf:
Python 3.11 is getting tomllib! 🎉 

https://docs.python.org/3.11/library/tomllib.html

Could use this when running on 3.11 or later. Unfortunately, will be a long time till pre-3.11 is deprecated and vendoring tomli will be irrelevant, but at least it's coming.

takluyver:
Thanks, this is indeed the plan. :slightly_smiling_face: Uses like this for packaging were a key part of the rationale for adding TOML parsing to the standard library.

nanonyme:
Is there any ETA on this? This might be useful for distros that take Python 3.11 into use next Fall.

takluyver:
No specific ETA, but it shouldn't be a big change. If anyone wants to make a PR, it should be a pretty easy way to become a Flit contributor. :slightly_smiling_face:

nanonyme:
I'm actually unsure on some reading how much this project would benefit of switching to tomllib. The intent is that you would even with tomllib still need tomli-w... which does depend on tomli. For flit_core bootstrapping tomllib would *kind of* make sense *but* the vendored tomli cannot be removed without dropping support for Python versions older than 3.11... So maybe this makes less sense than I first thought.

takluyver:
Yup, that sounds right. Anywhere we need to parse toml, we'd use tomllib with a fallback to tomli. It shouldn't make much difference until we can drop Python 3.10 support (which is some way off), except perhaps to downstream packagers who need to unbundle tomli.

PR #573, opened 2022-07-17
Use tomllib instead of vendored tomli when available

nanonyme:
This allows unvendoring tomli assuming Python 3.11 or newer. FIxes https://github.com/pypa/flit/issues/544


nanonyme:
So effectively what is done here is *only* flit_core switches to tomllib. tomli remains a test dependency and tests keep using tomli.

takluyver:
Thanks, this looks good.

PR #581, opened 2022-08-20
Force bytecode directories and files always to exclude

nanonyme:
https://github.com/pypa/flit/issues/568

### DIFF
--- a/flit_core/flit_core/config.py
+++ b/flit_core/flit_core/config.py
@@ -142,8 +142,12 @@ def prep_toml_config(d, path):
         loaded_cfg.sdist_include_patterns = _check_glob_patterns(
             dtool['sdist'].get('include', []), 'include'
         )
+        exclude = [
+            "**/__pycache__",
+            "**.pyc",
+        ] + dtool['sdist'].get('exclude', [])
         loaded_cfg.sdist_exclude_patterns = _check_glob_patterns(
-            dtool['sdist'].get('exclude', []), 'exclude'
+            exclude, 'exclude'
         )
 
     data_dir = dtool.get('external-data', {}).get('directory', None)
--- a/doc/pyproject_toml.rst
+++ b/doc/pyproject_toml.rst
@@ -420,7 +420,8 @@ These paths:
 - Should match the case of the files they refer to, as case-insensitive matching
   is platform dependent
 
-Exclusions have priority over inclusions.
+Exclusions have priority over inclusions. Bytecode is excluded by default and cannot
+be included.
 
 .. note::
 


nanonyme:
@takluyver is this what you suggested?

nanonyme:
I tried using this by byte-compiling flit_core, then creating sdist of flit_core. There were no pyc files inside said package.

flit_core/flit_core/config.py
```diff
@@ -142,8 +142,13 @@ def prep_toml_config(d, path):
         loaded_cfg.sdist_include_patterns = _check_glob_patterns(
             dtool['sdist'].get('include', []), 'include'
         )
+        exclude = [
+            "**/__pycache__",
+            "**.pyc",
+        ]
+        exclude.extend(dtool['sdist'].get('exclude', []))
```
takluyver, line 149:
```suggestion
        ] + dtool['sdist'].get('exclude', [])
```

takluyver:
Thanks, that is what I was suggesting. Can we add a note in the docs about the sdist section that these files are always excluded?

flit_core/flit_core/config.py
nanonyme: Hmm, extend avoids creating unnecessary copy of the list which happens when + is used. I can change if desired.

nanonyme:
Sure, I will apply requested changes later today.

flit_core/flit_core/config.py
takluyver: I prefer the clearer code over micro-optimisation. I'm not sure the `.extend()` version is necessarily faster anyway, but even if it is, the difference is going to be meaningless compared to everything else we're doing.

nanonyme: What the proposed code does is it will create a list, throw it away and create a new list with contents of both parts. Extend will append into existing list and is no-op when appended item is empty. It is always faster to extend. But ok, if you feel it makes code clearer.

takluyver: It's not *always* faster to extend:

```
In [3]: %%timeit
   ...: a = [1, 2] + [b, 4, 5]
149 ns ± 2.52 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)

In [4]: %%timeit
   ...: a = [1, 2]
   ...: a.extend([b, 4, 5])
201 ns ± 18.1 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
```

This level of performance is really hard to guess about. Is creating new objects what we should focus on? Or bytecode instructions? Does a method call have more or less overhead than an operator? When does a list reallocate while growing, and is either option smart enough to allocate the final size up front? How does it change from one version of Python to the next?

But any possible difference is so tiny that it's not worth figuring this out. Premature optimisation is the root of all evil. :wink: Write clear, working code, and if we need it to be faster, work out which bits are taking the most time (which almost certainly won't be this bit).

takluyver:
Thanks!

PR #586, opened 2022-08-25
Enable `~=` syntax when checking Py2 compatibility

### LINKED ISSUE #476: `python-requires="~3.7"` generates a universal Wheel file
When this is in `pyproject.toml`:

    [project]
    # ...
    requires-python = ">=3.7,<4"

The `flit build` produces a Python-3-only Wheel file: `projectname-0.0.1.dev0-py3-none-any.whl`.  Good! (and thanks for Flit, it's really nice!)

When I use the shorter syntax to require Python 3.7 or higher -- which I understood to be equivalent:

    requires-python = "~=3.7"

Then the produced Wheel becomes universal: `projectname-0.0.1.dev0-py2.py3-none-any`.  
When - obviously - I don't want to support Python 2.

If this difference in behavior is intentional, it should at least be mentioned in [the docs](https://flit.readthedocs.io/en/latest/pyproject_toml.html#new-style-metadata).

LEdoian:
This should fix #476.

I do not remember full [PEP 440](https://peps.python.org/pep-0440/), so I am not sure whether there are other valid expressions to disallow Python 2 which would still be false-positive in flit.

Probably it would be better to use [specifiers from `packaging`](https://packaging.pypa.io/en/latest/specifiers.html) library so that flit does not need to implement own specifier parsing. Therefore this is just a quick fix for the issue at hand, which makes a small improvement but probably is not final.

### DIFF
--- a/flit_core/flit_core/tests/test_common.py
+++ b/flit_core/flit_core/tests/test_common.py
@@ -113,6 +113,8 @@ def test_normalize_file_permissions():
         ("<4, > 3.2", False),
         (">3.4", False),
         (">=2.7, !=3.0.*, !=3.1.*, !=3.2.*", True),
+        ("~3.3", False),
+        ("~2.7", True),
     ],
 )
 def test_supports_py2(requires_python, expected_result):
--- a/flit_core/flit_core/common.py
+++ b/flit_core/flit_core/common.py
@@ -398,7 +398,7 @@ class Metadata(object):
     def supports_py2(self):
         """Return True if Requires-Python indicates Python 2 support."""
         for part in (self.requires_python or "").split(","):
-            if re.search(r"^\s*(>\s*(=\s*)?)?[3-9]", part):
+            if re.search(r"^\s*([>~]\s*(=\s*)?)?[3-9]", part):
                 return False
         return True
 


takluyver:
Thanks! Can you add a case or two to the test for this:

https://github.com/pypa/flit/blob/f86127764bddb05cb6b5d9a1b6ae4346f786c94a/flit_core/flit_core/tests/test_common.py#L106-L122

nanonyme:
@LEdoian flit cannot depend on packaging because packaging needs (or will need if everything goes as expected) flit to build.

LEdoian:
@takluyver: Done.

LEdoian:
@nanonyme: That seems unfortunate, but I understand this is not worth complicating the bootstrapping processes.

nanonyme:
It wouldn't just complicate it but make it impossible. This package is at the very bottom of the food chain, it cannot have any external dependencies.

LEdoian:
I do not think it would be outright impossible, but I do not have a simple and elegant solution.

The idea would be to try importing `packaging`. If it succeeds, use the full specifier matching, if not, fall back to the current regex one with its limitations. This way `packaging` could be built and everyone else can use full-fledged version (since once `packaging` is built, it does not need flit anymore).

But this would be needlessly complex and awful from software engineering standpoint, especially given that flit tries to be the simple option. So not worth it.

flit_core/flit_core/tests/test_common.py
```diff
@@ -113,6 +113,8 @@ def test_normalize_file_permissions():
         ("<4, > 3.2", False),
         (">3.4", False),
         (">=2.7, !=3.0.*, !=3.1.*, !=3.2.*", True),
+        ("~3.3", False),
+        ("~2.7", True),
```
takluyver, line 117: `~3.3` is not actually a valid version specifier according to [PEP 440](https://peps.python.org/pep-0440/#version-specifiers), it should be `~=`.

I think we should make the different supported operators explicit, i.e. `(>=?|~=|==)`.

takluyver:
And yes, I agree with you that trying the import and falling back to a simpler implementation is not desirable - it's extra complexity, and people will still want to fix the bugs with the fallback option.

I think doing a subset of proper version specifier handling inside flit_core is feasible, because a) it only needs to handle Python's predictable X.Y.Z version numbers (I checked the code in pip, that ignores anything beyond the three numbers, so we can do the same), and b) we don't need the nice `SpecifierSet` API that packaging offers.

takluyver:
(To be clear, I'm not saying you need to do that in this PR, just thinking about the direction we might go in)

takluyver:
I've just merged #593, which solved the same thing.

Issue #591, opened 2022-09-06
wheel name not matching requirements specifier.


dougransom:
command `flit build --format wheel`

Try requirements specifications:
`requires-python="== 3.9"`
`Built wheel: dist\natlink-5.3.2-py2.py3-none-any.whl I-flit_core.wheel`
same with 
`requires-python="3.9.*"`
`requires-python="~= 3.1"`

was expecting natlink-5.3.2-py3.9-none-any.whl  or something similar without py2.

takluyver:
The version checking that decides whether to add the `py2` tag is a crude bit of code which has survived since basically the first draft of Flit, and has numerous bugs (see e.g. #586). Really, we should properly parse it and check with '2.7' using the rules described in [PEP 440](https://peps.python.org/pep-0440/#version-specifiers) - but it's always been easier to simply tweak the regex a bit to solve the most visible problems.

Note that `requires-python="3.9.*"` is not strictly valid - the value should always start with one of the version comparison operators (`> >= < <= == != ~= ===`). `flit build` should complain about this, but using flit_core as a backend to other tools won't, because that assumes the perspective of a user who wants to install something whether or not the metadata is correct.

dougransom:
Can flit be dependent on [packaging](https://packaging.pypa.io/en/latest/index.html)?

takluyver:
`flit_core`, where the relevant functionality needs to be, can't really use packaging, because we're aiming for flit_core to be the starting point for [bootstrapping](https://flit.pypa.io/en/stable/bootstrap.html) the ecosystem of packaging tools.

I'd hope that we can have a significantly simpler implementation than packaging, because we don't need a nice API for it, and we only need to handle the reliable version numbers of Python itself, not the full variety of possible version numbers out there. I found the code in pip that checks the requires-python specifier, and it looks like it only uses the three simple numeric parts of the version (e.g. 3.10.6), so it should be OK for us to do the same:

https://github.com/pypa/pip/blob/0d4e9eb72253c008f2790482e664ce92198c5240/src/pip/_internal/utils/packaging.py#L14

PR #593, opened 2022-09-19
Modify regex for checking py2 compatibility

### LINKED ISSUE #476: `python-requires="~3.7"` generates a universal Wheel file

EwaMarek:
Closes #476 and #591
Add support for compatible release version specifier '~=' and version matching clause '=='.

I was happy to see that someone picked up this issue in #586 , but it seems like the conversation stalled a bit, so here I am. It's something that was bothering me for a while now and it would be nice to see it fixed 😄

### DIFF
--- a/flit_core/flit_core/common.py
+++ b/flit_core/flit_core/common.py
@@ -398,7 +398,7 @@ class Metadata(object):
     def supports_py2(self):
         """Return True if Requires-Python indicates Python 2 support."""
         for part in (self.requires_python or "").split(","):
-            if re.search(r"^\s*(>\s*(=\s*)?)?[3-9]", part):
+            if re.search(r"^\s*(>=?|~=|===?)?\s*?[3-9]", part):
                 return False
         return True
 
--- a/flit_core/flit_core/common.py
+++ b/flit_core/flit_core/common.py
@@ -398,7 +398,7 @@ class Metadata(object):
     def supports_py2(self):
         """Return True if Requires-Python indicates Python 2 support."""
         for part in (self.requires_python or "").split(","):
-            if re.search(r"^\s*(>=?|~=|===?)?\s*?[3-9]", part):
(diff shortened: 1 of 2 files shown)


flit_core/flit_core/common.py
```diff
@@ -398,7 +398,7 @@ def write_metadata_file(self, fp):
     def supports_py2(self):
         """Return True if Requires-Python indicates Python 2 support."""
         for part in (self.requires_python or "").split(","):
-            if re.search(r"^\s*(>\s*(=\s*)?)?[3-9]", part):
+            if re.search(r"^\s*(>=?|~=|===?)?\s*?[3-9]", part):
```
joukewitteveen, line 401:
```suggestion
            if re.search(r"^\s*(>=?|~=|===?)?\s*[3-9]", part):
```
Was there a reason to have the `?` following `\s*`? I don't think it achieves anything ("zero or one times zero or more spaces").

EwaMarek: Right, this `?` doesn't achieve much, thanks for spotting it!

takluyver:
Thanks, this looks good. :slightly_smiling_face:

PR #601, opened 2022-10-03
docs: drop toml dependency from development.rst

joukewitteveen:
toml has been replaced by tomli and tomli-w. The latter is not needed to get a development installation and the former is either available as a standard library (as tomllib), or vendored from flit itself. Either way, no toml-related installation is necessary.

Inspired by a suggestion by Joe Drumgoole <Joe.Drumgoole@mongodb.com>

---
This is a follow-up on #480 :smiley:.


takluyver:
Thanks!

Issue #276, opened 2019-07-22
support token-based upload to PyPI


brainwane:
Hi! PyPI's getting ready to add upload tokens to its API, so packagers can upload using tokens instead of username/password.

I thought you might want to test and verify that flit's upload works with this. Instructions for testing the pull request: https://github.com/pypa/warehouse/issues/994#issuecomment-512634222

It would be great if you could leave a comment on https://github.com/pypa/warehouse/pull/6084 telling us whether it works for you! This will help us avoid headaches when we roll out the feature on pypi.org.

takluyver:
Thanks. I'm just about to go offline for a week, and then it will be another few days while I dig myself out from the emails that build up. But I'll try to get round to it.

brainwane:
Please do enjoy your offline time! And if someone else would like to do the testing -- maybe @flying-sheep? -- that would, of course, be fine!

brainwane:
This beta feature is now live on PyPI and Test PyPI: https://pypi.org/help/#apitoken

Carreau:
Tried it; worg great. Just set `FLIT_USERNAME='@token'` and `FLIT_PASSWORD='<yourtoken>'`. 

One question we may want to adress is how we pick [which token for which project](https://github.com/pypa/packaging.python.org/issues/628) if not in an env variable.

takluyver:
Brainstorming how the UI could work:

If I'm trying to publish a release of `austin` and my `~/.pypirc` says:

```ini
[pypi]
repository: https://upload.pypi.org/legacy/
username: takowl
```

Then flit could use the first it finds of:

```python
# Project scoped token
keyring.get_password("https://upload.pypi.org/legacy/", "pypi-project-token:austin")

# User scoped token
keyring.get_password("https://upload.pypi.org/legacy/", "pypi-user-token:takowl")
```

If it doesn't find either of those, it would ideally connect as my user, prompting me for any necessary credentials, and then obtain and save a project-scoped token to use for uploads. If that's not possible, it might fall back to uploading with username & password as it currently does.

Does that scheme fit in with how you'd like to use tokens? Of course, it could all be overridden by env variables for CI - I'm just thinking about the upload-from-local case.

brainwane:
cc @di and @ewdurbin and @dstufft to answer this question. @woodruffw as an FYI.

Carreau:
[...]

PR #602, opened 2022-10-04
Document that token-based upload is supported

Mariatta:
Closes https://github.com/pypa/flit/issues/276

### DIFF
--- a/doc/cmdline.rst
+++ b/doc/cmdline.rst
@@ -207,6 +207,10 @@ Environment variables
    Set a username, password, and index URL for uploading packages.
    See :ref:`uploading packages with environment variables <upload_envvars>`
    for more information.
+   
+   Token-based upload to PyPI is supported. To upload using PyPI token,
+   set the ``FLIT_USERNAME`` value to ``__token__``, and the ``FLIT_PASSWORD``
+   to the token value.
 
 .. envvar:: FLIT_ALLOW_INVALID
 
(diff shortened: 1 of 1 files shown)


pradyunsg (approved):
Thanks @Mariatta!

doc/cmdline.rst
```diff
@@ -207,6 +207,10 @@ Environment variables
    Set a username, password, and index URL for uploading packages.
    See :ref:`uploading packages with environment variables <upload_envvars>`
    for more information.
+   
+   Token-based upload to PyPI is supported. To upload using PyPI token,
+   set the ``FLIT_USERNAME`` value to ``__token__``, and the ``FLIT_PASSWORD``
+   to the token value.
```
joukewitteveen, line 213:
```suggestion
   Token-based upload to PyPI is supported. To upload using a PyPI token,
   set ``FLIT_USERNAME`` to ``__token__``, and ``FLIT_PASSWORD`` to the
   token value.
```

joukewitteveen (commented):
Not sure if this needs to be in the flit documentation, since it is not specific to flit.

Mariatta:
I was publishing using flit, and it was not obvious to me that I could use the PyPI token instead of regular username/password. Additionally, while searching for "upload using token with flit" I came across this open issue on GitHub which made me think that it was not supported. So I thought that it would be better to explicitly document it.

pradyunsg:
To err on the side of overcommunicating, I'm unlikely to be merging this PR.

I have a green tick cos I'm an admin on pypa but I'm not a committer to twine, and would prefer that someone who is lands this instead. :)

takluyver:
Thanks @Mariatta !

@pradyunsg FWIW, I'm happy for you to merge something like this if you're comfortable doing so.

PR #603, opened 2022-10-07
Include license file in the built package

gotmax23:
As far as I understand (IANAL), distributing the package on PyPI without the license text violates your own license terms.


takluyver:
Thanks! I'm not sure about the legalities, but it's certainly a good idea to have the license in there.

PR #604, opened 2022-10-16
Remove vestigial tomli usage 

gotmax23:
dba0f317c52d3c9fa29792a0beb0c0abf4476a23 removed some, but not all of the imports of the vendored tomli. This PR adds the same tomllib > tomli fallback to the tests and adjusts the tox.ini test deps accordingly.

Additionally, this adjusts flit-core to only use its vendored tomli when tomli isn't already installed. In Fedora, we bootstrap tomli differently, so we `rm -r` the vendored version. This change makes it so Fedora and other distributions don't also have to patch flit_core.config.

/cc @hroncok

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       matrix:
         platform: ["ubuntu-latest", "windows-latest"]
-        python-version: [ "3.6", "3.7", "3.8", "3.9", "3.10" ]
+        python-version: [ "3.6", "3.7", "3.8", "3.9", "3.10", "3.11" ]
     steps:
       - uses: actions/checkout@v3
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,7 @@ jobs:
     strategy:
       matrix:
         platform: ["ubuntu-latest", "windows-latest"]
-        python-version: [ "3.6", "3.7", "3.8", "3.9", "3.10", "3.11" ]
+        python-version: [ "3.6", "3.7", "3.8", "3.9", "3.10", "3.11-dev" ]
     steps:
       - uses: actions/checkout@v3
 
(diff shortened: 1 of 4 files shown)


gotmax23:
The tests don't currently run on Python 3.11, which makes this not particularly useful upstream. Would you like me to also add Python 3.11 to the test matrix as part of this change?

hroncok:
Thanks.

flit_core/flit_core/config.py
```diff
@@ -10,7 +10,10 @@
 try:
     import tomllib
 except ImportError:
-    from .vendor import tomli as tomllib
+    try:
+        import tomli as tomllib
+    except ImportError:
+        from .vendor import tomli as tomllib
```
takluyver, line 16: Is it OK for you if we swap round the logic here - i.e. try the vendored version first, and only load a separate tomli if the vendored one is removed?

My reason for preferring this is that it means there's no difference in behaviour between when tomli is installed or not.

gotmax23:
@takluyver, can you please approve the workflow for this PR and #603?

flit_core/flit_core/config.py
gotmax23: Yes. I understand the reasoning, and that works fine for us.

.github/workflows/test.yml
```diff
@@ -19,7 +19,7 @@ jobs:
     strategy:
       matrix:
         platform: ["ubuntu-latest", "windows-latest"]
-        python-version: [ "3.6", "3.7", "3.8", "3.9", "3.10" ]
+        python-version: [ "3.6", "3.7", "3.8", "3.9", "3.10", "3.11" ]
```
takluyver, line 22:
```suggestion
        python-version: [ "3.6", "3.7", "3.8", "3.9", "3.10", "3.11-dev" ]
```

Looks like we need this until 3.11 is actually released. I'll commit this change myself to test it quickly.

hroncok:
Thank you!

nanonyme:
Something I'm wondering if it could be unvendored for install and vendored only for the purposes of sdist and repo. And we instead add vendored tomli through backend-path.

PR #615, opened 2022-12-16
Use Ubuntu 20.04 to test on Python 3.6

takluyver:
`ubuntu-latest` now refers to 22.04, which doesn't have 3.6 available.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -19,7 +19,31 @@ jobs:
     strategy:
       matrix:
         platform: ["ubuntu-latest", "windows-latest"]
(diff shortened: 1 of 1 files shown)


takluyver:
I'll merge this one with no review, since it's a simple change just to get CI working again.

takluyver:
It's probably reasonable to next time there's a decent reason, but I'm not in a hurry. We've only fairly recently dropped 3.6 for some projects at work, and I think low-level packaging tools should be fairly conservative about supporting older Pythons.

I'm also OK with `flit` requiring a newer Python than `flit_core` - this was the case in the past, and it's kind of coincidence that they have the same minimum version now. But that would still mean running (some of) the tests.

PR #621, opened 2023-01-12
Add more robust license file installation

gotmax23:
- Expand default license file inclusion globs. This uses the same patterns as setuptools/wheel with the addition of 'LICENSES/*.txt' for REUSE.
- Ensure matching license files are always included in sdists.
- Warn users when no license files are detected.
- Add unit tests

### DIFF
--- a/flit_core/flit_core/config.py
+++ b/flit_core/flit_core/config.py
@@ -21,6 +21,9 @@ from .versionno import normalise_version
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


AA-Turner:

A

takluyver:
Sorry I didn't get round to looking at this - the plan was probably fine, but I don't have a ton of time to work on Flit, and it's hard to get interested in stuff around license files. Especially when PEP 639 was under discussion so long, so I knew I'd be coming back to the topic when that was accepted.

PR #622, opened 2023-01-23
Fix typo

cclauss:


Mariatta:
The failed test is not caused by this change, but due to deprecated option in tox library.
#624 should fix that.

Mariatta (approved):
`inferred` is correct. Thanks for fixing.

takluyver:
I'll just do a quick close/reopen to let the tests run with #624 merged, but this looks good, thanks.

PR #623, opened 2023-01-23
Test on the production release of Python 3.11

cclauss:
And the dev release of Python 3.12


Mariatta:
The failed test is not caused by this change, but due to deprecated option in tox library.
#624 should fix that.

cclauss:
The credit goes to https://pypi.org/project/codespell

takluyver:
Thanks! As with #622, I'll close and reopen to re-run the tests with a new merge.

PR #624, opened 2023-01-24
tox: use allowlist_externals option instead of whitelist_externals

Mariatta:
`whitelist_externals` was removed and deprecated in tox v 4.0.0.rc4 We should use `allowlist_externals` instead.

https://tox.wiki/en/latest/changelog.html#deprecations-and-removals-4-0-0rc4


takluyver:
Thanks!

Issue #522, opened 2022-02-02
Discussion: future shape of the 'flit' command


takluyver:
Flit came about in a very different world, before various standards defined how packaging tools should work together (in particular, PEPs [517](https://www.python.org/dev/peps/pep-0517/), [621](https://www.python.org/dev/peps/pep-0621/) and [660](https://www.python.org/dev/peps/pep-0660/)). A lot of design decisions I took back then don't really fit with the world we have now, and I don't have a strong sense of how they should be reconciled. So, let's have a discussion. :slightly_smiling_face: 

First, a couple of specific examples of where Flit doesn't fit in:

- Flit works as a PEP 660 backend for editable installs, creating `.pth` files to add entries to `sys.path`. But I happen to prefer doing editable installs with symlinks, which was implemented in Flit long before there was a standardised API. So if you use the Flit command, this is still possible as `flit install --symlink`. A fair bit of code which was more general now basically remains just for this feature. See also #512.
- `flit build` creates an sdist using information from git (or mercurial) to decide which files to include. I envisaged that you would only build sdists to upload to PyPI, and it was reasonable to assume you did that from a git checkout with the `git` command available. But PEP 517 made `build_sdist` a standard hook, and I didn't want to make the same assumption in the PEP 517 backend. So `flit build` gives you what I think of as the 'gold standard' sdist, but using e.g. `python -m build`, which calls the PEP 517 hook, gives you a 'minimal' sdist unless you add include/exclude patterns in your config file. See also #513.

But beyond the specifics, I want to ask a more general question about the role of Flit. Part of my goal was for Flit to provide a single CLI for tasks around making & sharing a package, similar to how pip is a single CLI for consuming packages. `flit init` helps you set up the metadata, `flit install` lets you try it out, and `flit publish` checks everything and uploads it to PyPI. I still use it that way - old habits die hard - but with the new wave of standards, people often recommend using tools that work with any backend - `pip` to install, `build` to, uh, build, `twine` to upload. It's a worse experience if you're working on one particular package, but it works the same way for any package.

So, what do we want? I see 3 main options:

1. Gradually deprecate most (or all) of the Flit command, pushing people towards pip/build/twine to take its place. What's left is basically the simplest practical packaging backend (and thus hopefully reliable, easy to understand, a good place to try implementing new standards...)
2. Keep the `flit` command around as an alternative interface for people (like me!) who want to use it. The status quo isn't actually broken, after all - though I'd like to clean up things like how to decide what files to put in an sdist.
[...]

PR #625, opened 2023-01-27
Add --use-vcs and --no-use-vcs flags for build commands

takluyver:
From the discussion on #522, this adds an option to not use the VCS to get the list of files to include in the sdist. So `flit build --no-use-vcs` is equivalent to `python -m build` with flit_core as a backend. The idea is that in a future version (4.0?), this will become the default, and you will need to use `--use-vcs` to explicitly choose the sdist-files-from-vcs behaviour.

I've also tried to improve the documentation about what gets included in the sdist by default.

### DIFF
--- a/doc/pyproject_toml.rst
+++ b/doc/pyproject_toml.rst
@@ -400,10 +400,13 @@ Sdist section
 
 .. versionadded:: 2.0
 
-When you use :ref:`build_cmd` or :ref:`publish_cmd`, Flit builds an sdist
-(source distribution) tarball containing the files that are checked into version
-control (git or mercurial). If you want more control, or it doesn't recognise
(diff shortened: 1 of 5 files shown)


pfmoore:
+1 from me. The docs changes look good, and the plan for switching the default seems reasonable. I can't really comment on the code so I'll leave that for others.

One thing that I'd like which is missing, is a clear description of what files will be put in the *wheel*. Is it just "the stuff in the sdist that's from the package directory"? Having it explicitly stated would avoid any guessing.

takluyver:
@Carreau @pradyunsg I'm planning to merge this in a couple more days, so if you're not happy with this direction, now's the time to say so. :slightly_smiling_face:

Carreau (approved):
+1 from me.

PR #628, opened 2023-02-27
flit_core: Normalize sdist names according to PEP 625

### LINKED ISSUE #626: flit-core: Implement sdist name normalization (PEP 625)
https://packaging.python.org/en/latest/specifications/source-distribution-format/#source-distribution-file-name
https://peps.python.org/pep-0625/

Basically, the distribution name and version needs to be normalized the same way as for wheels.

mgorny:
Normalize generated source distribution names using the same rules as wheels, as required by PEP 625 [1].  This aligns flit_core with other modern PEP 517 backends such as hatchling.

[1] https://peps.python.org/pep-0625/

Fixes #626

### DIFF
--- a/flit_core/flit_core/sdist.py
+++ b/flit_core/flit_core/sdist.py
@@ -157,13 +157,11 @@ class SdistBuilder:
 
     @property
     def dir_name(self):
-        return '{}-{}'.format(self.metadata.name, self.metadata.version)
+        return common.normalize_dist_name(self.metadata.name, self.metadata.version)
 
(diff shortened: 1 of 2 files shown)


mgorny:
I kinda suspect that it was too easy ;-).

takluyver:
Thanks! I know what you mean that it seems too easy, but I can't see anything wrong, so :crossed_fingers: it's OK. Sorry it's taken me a while to look at it.

mgorny:
Thanks!

PR #630, opened 2023-03-12
Read version from version files via ast

amyreese:
In cases where packages import `__version__` from version modules (like
those generated by setuptools_scm or attribution), this allows flit-core
to use the existing ast-based logic to read the version directly from
those files when possible. It checks a few well-known version filenames,
including the package-level `__init__.py` as usual, with `__init__.py`
taking precedence over external version files if an assigned
`__version__` value exists in multiple places.

This makes flit more resilient to reading version from these packages,
even if their `__init__.py` has transitive imports that aren't available
in the environment at build time (eg, conda-forge builds). It also
reduces the number of times when flit has to fall back to actually
importing the `__init__.py` file rather than reading with ast.

### DIFF
--- a/flit_core/flit_core/common.py
+++ b/flit_core/flit_core/common.py
@@ -71,6 +71,17 @@ class Module(object):
         else:
             return self.path
 
+    @property
+    def version_files(self):
+        if self.is_package:
+            paths = [self.path / '__init__.py']
+            for filename in ('version.py', '_version.py', '__version__.py'):
+                if (self.path / filename).is_file():
+                    paths.insert(0, self.path / filename)
+            return paths
+        else:
+            return [self.path]
+
     def iter_files(self):
         """Iterate over the files contained in this module.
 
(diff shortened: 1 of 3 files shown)


amyreese:
Being able to read `__version__` from a separate file, without needing to import the entire package, would resolve omnilib/ufmt#123 and mitigate the need for conda-forge to install runtime dependencies at build time to support packages that import `__version__` from a separate file.

akx:
FWIW, there's prior art in pypa/setuptools for this: https://github.com/pypa/setuptools/blob/be6c0218bcba78dbd4ea0b5a8bb9acd5d5306240/setuptools/config/expand.py#L160-L168

flit_core/flit_core/common.py
```diff
@@ -127,26 +138,26 @@ def get_docstring_and_version_via_ast(target):
...
+                        isinstance(target, ast.Name)
+                        and target.id == "__version__"
+                        for target in child.targets
+                    )
+                    and isinstance(child.value, ast.Str)
+            )
+            if is_version_str:
+                version = child.value.s
+                break
     return ast.get_docstring(node), version
```
takluyver, line 161: I think the docstring should always come from the main file (i.e. `__init__.py` if this is a package), rather than the version file. The first line of this docstring is used as the short description for PyPI by default.

takluyver:
Thanks! There's one little change I want (comment above), but I like the idea and the implementation. :+1:

flit_core/flit_core/common.py
amyreese: @takluyver I believe that's how this works as-implemented. The order of lookup is intentionally crafted with the `__init__.py` as the last entry `target.version_files`, and the loop doesn't short circuit. The docstring lookup only happens after the loop is completed, when `node` will always have the contents of `__init__.py`.

The existing test case [1][] passes, and contains the docstring from the fixture's `__init__.py`. I went ahead and pushed an additional change to add an "imposter" docstring to the `_version.py` fixture, to help show that it's not being used/selected.

[1]: https://github.com/pypa/flit/blob/fc640d2c79441bc8a37e5098db898ecd42bcc44f/flit_core/flit_core/tests/test_common.py#L83

takluyver: You're right, thanks. I had scanned the code and assumed the `break` would stop at the first file containing `__version__`, but of course it only breaks out of the inner loop.

flit_core/flit_core/common.py
```diff
@@ -71,6 +71,17 @@ def file(self):
         else:
             return self.path
 
+    @property
+    def version_files(self):
```
takluyver, line 75:
```suggestion
    def version_files(self):
    """Files which will be parsed to find a version number
    
    Files later in this list take precedence over earlier ones.
    """
```

takluyver:
Thanks @amyreese, and nice work on a first contribution. :slightly_smiling_face:

PR #631, opened 2023-03-12
Enable --only-deps with FLIT_ALLOW_INVALID

dciborow:
Currently, trying to run "flit install --only-deps" includes checks that require both the README and the module folder be present, even though they are not actually needed. When trying to use --only-deps in my docker file, I have to also copy the README and have to create a fake directory.

```Dockerfile
FROM mcr.microsoft.com/devcontainers/python:3

RUN python -m pip install --upgrade pip \
    && python -m pip install pytest pytest-cov \
    && python -m pip install 'flit>=3.8.0'

ENV FLIT_ROOT_INSTALL=1

COPY pyproject.toml README.rst ./
RUN mkdir -p flit \
    && python -m flit install --only-deps --deps develop \
    && rm -r pyproject.toml README.rst flit
```

flit already includes a flag to try and install if there are invalid settings, `FLIT_ALLOW_INVALID`. I have extended the use of that flag to enable running --only-deps without the checks. The updated Dockerfile is.

```Dockerfile
FROM mcr.microsoft.com/devcontainers/python:3

RUN python -m pip install --upgrade pip \
    && python -m pip install pytest pytest-cov \
    && python -m pip install 'flit>=3.8.1'

ENV FLIT_ROOT_INSTALL=1
ENV FLIT_ALLOW_INVALID=1

COPY pyproject.toml .
RUN python -m flit install --only-deps --deps develop \
    && rm -r pyproject.toml
```

### DIFF
--- a/flit_core/flit_core/config.py
+++ b/flit_core/flit_core/config.py
@@ -25,137 +25,142 @@ log = logging.getLogger(__name__)
 class ConfigError(ValueError):
     pass
 
-metadata_list_fields = {
-    'classifiers',
-    'requires',
-    'dev-requires'
-}
+
+metadata_list_fields = {"classifiers", "requires", "dev-requires"}
 
 metadata_allowed_fields = {
-    'module',
-    'author',
-    'author-email',
-    'maintainer',
-    'maintainer-email',
-    'home-page',
-    'license',
-    'keywords',
-    'requires-python',
-    'dist-name',
-    'description-file',
-    'requires-extra',
+    "module",
+    "author",
+    "author-email",
+    "maintainer",
+    "maintainer-email",
+    "home-page",
+    "license",
+    "keywords",
+    "requires-python",
+    "dist-name",
+    "description-file",
+    "requires-extra",
 } | metadata_list_fields
 
 metadata_required_fields = {
-    'module',
-    'author',
+    "module",
+    "author",
 }
 
 pep621_allowed_fields = {
-    'name',
-    'version',
-    'description',
-    'readme',
-    'requires-python',
-    'license',
-    'authors',
-    'maintainers',
-    'keywords',
-    'classifiers',
-    'urls',
-    'scripts',
-    'gui-scripts',
-    'entry-points',
-    'dependencies',
-    'optional-dependencies',
-    'dynamic',
+    "name",
+    "version",
+    "description",
+    "readme",
+    "requires-python",
+    "license",
+    "authors",
+    "maintainers",
+    "keywords",
+    "classifiers",
+    "urls",
+    "scripts",
+    "gui-scripts",
+    "entry-points",
+    "dependencies",
+    "optional-dependencies",
+    "dynamic",
 }
 
 
 def read_flit_config(path):
-    """Read and check the `pyproject.toml` file with data about the package.
-    """
-    d = tomllib.loads(path.read_text('utf-8'))
+    """Read and check the `pyproject.toml` file with data about the package."""
+    d = tomllib.loads(path.read_text("utf-8"))
     return prep_toml_config(d, path)
 
 
 class EntryPointsConflict(ConfigError):
     def __str__(self):
-        return ('Please specify console_scripts entry points, or [scripts] in '
-            'flit config, not both.')
+        return (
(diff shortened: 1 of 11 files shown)


dciborow:
@takluyver , have made some updates so that I can use `FLIT_ALLOW_INVALID` to run `flit install --only-deps` with just a pyproject.toml (particularly for building Docker containers)

flit_core/flit_core/config.py
```diff
@@ -304,11 +304,11 @@ def description_from_file(rel_path: str, proj_dir: Path, guess_mimetype=True):
 
 def _prep_metadata(md_sect, path):
     """Process & verify the metadata from a config file
-    
+
     - Pull out the module name we're packaging.
     - Read description-file and check that it's valid rst
     - Convert dashes in key names to underscores
-      (e.g. home-page in config -> home_page in metadata) 
+      (e.g. home-page in config -> home_page in metadata)
```
dciborow, line 311:
```suggestion
      (e.g. home-page in config -> home_page in metadata) 
```

tests/samples/only-deps.toml
```diff
@@ -0,0 +1,10 @@
+[build-system]
+requires = ["flit"]
+
+[tool.flit.metadata]
+module = "nomodule"
+author = "Sir Robin"
+author-email = "robin@camelot.uk"
+home-page = "http://github.com/sirrobin/module1"
+description-file = "NO_README.rst"
```
dciborow, line 9:
```suggestion
description-file = "definitely-missing.rst"
```

flit_core/flit_core/config.py
```diff
@@ -281,6 +281,8 @@ def description_from_file(rel_path: str, proj_dir: Path, guess_mimetype=True):
         with desc_path.open('r', encoding='utf-8') as f:
             raw_desc = f.read()
     except IOError as e:
+        if os.environ.get('FLIT_ALLOW_INVALID'):
+            return None, None
```
dciborow, line 285:
```suggestion
            log.warning("Allowing invalid data (FLIT_ALLOW_INVALID set). Uploads may still fail.")
            return None, None
```

tests/test_install.py
```diff
@@ -278,6 +279,32 @@ def test_install_requires(self):
...
+            assert calls[0]['argv'][1:5] == ['-m', 'pip', 'install', '-r']
+        finally:
+            del os.environ['FLIT_ALLOW_INVALID']
+
+    def test_install_only_deps_fail(self):
+        with pytest.raises(ConfigError, match=r"Description file .* does not exist"):
+            Installer.from_ini_path(
+                samples_dir / 'missing-description-file.toml', user=False, python='mock_python'
+            )
+        with pytest.raises(ValueError, match=r"No file/folder found for module nomodule"):
```
dciborow, line 303:
```suggestion
        with pytest.raises(ValueError, match=r"No file/folder found for module definitelymissingmodule"):
```

takluyver:
Hi @dciborow, sorry it's taken me a while to look at this.

I see what you're after, but this feels like the wrong approach to me. Flit is meant to work with a package, and crucial files being missing is not the same thing as metadata that Flit thinks is invalid. The FLIT_ALLOW_INVALID environment variable is basically meant as an 'escape hatch' for bugs in Flit's validation, which is why [the docs say](https://flit.pypa.io/en/stable/cmdline.html#envvar-FLIT_ALLOW_INVALID):

> If you need to use it and you believe your metadata is valid, please [open an issue](https://github.com/pypa/flit/issues).

Maybe `--only-deps` should use a separate code path which doesn't look at these things in the first place. Maybe it should even be a separate tool, not tied to Flit - it could be just a few lines, reading requirements in [the standard format](https://packaging.python.org/en/latest/specifications/declaring-project-metadata/#dependencies-optional-dependencies) from `pyproject.toml` and passing them to pip. Perhaps it should even be part of pip itself - like it can read a `requirements.txt` file, perhaps it should also allow a list of requirements from TOML. :thinking: 

@pradyunsg do you have any inspiration here? :slightly_smiling_face:

pradyunsg:
There's plans to add `--only-deps` to pip.

https://github.com/pypa/pip/issues/11440

PR #632, opened 2023-03-12
Add devcontainer configuration

dciborow:
Adding a devcontainer configuration which can be used with GitHub Codespace or with VSCode to create a local container.

The configuration includes the basics configurations for Python.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,12 @@
+FROM mcr.microsoft.com/devcontainers/python:3
+
+RUN python -m pip install --upgrade pip \
+    && python -m pip install pytest pytest-cov \
+    && python -m pip install 'flit>=3.8.0'
+
+ENV FLIT_ROOT_INSTALL=1
+
+COPY pyproject.toml README.rst ./
+RUN mkdir -p flit \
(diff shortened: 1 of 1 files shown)


.devcontainer/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+COPY pyproject.toml README.rst ./
+RUN mkdir -p flit \
+    && python -m flit install --only-deps --deps develop \
+    && rm -r pyproject.toml README.rst flit
+
+// ENV FLIT_ALLOW_INVALID=1
+
+// COPY pyproject.toml .
+// RUN python -m flit install --only-deps --deps develop \
+//     && rm -r pyproject.toml
```
dciborow, line 18:
```suggestion
    && python -m pip install 'flit>=3.8.1'

ENV FLIT_ROOT_INSTALL=1
ENV FLIT_ALLOW_INVALID=1

COPY pyproject.toml .
RUN python -m flit install --only-deps --deps develop \
    && rm -r pyproject.toml
```

dciborow: after this PR is completed, we can update the docker file.
https://github.com/pypa/flit/pull/631

PR #639, opened 2023-04-20
fix: regular flag was not set

### LINKED ISSUE #634: RECORD has incorrect file mode in generated archives
The file mode is incorrect for generated files:

```console
$ zipinfo dist/*.whl
...
?rw-r--r--  2.0 unx      104 b- defN 16-Jan-01 00:00 wheel-0.38.4.dist-info/entry_points.txt
-rw-r--r--  2.0 unx     1107 b- defN 23-Jan-27 13:33 wheel-0.38.4.dist-info/LICENSE.txt
?rw-r--r--  2.0 unx       81 b- defN 16-Jan-01 00:00 wheel-0.38.4.dist-info/WHEEL
?rw-r--r--  2.0 unx     2051 b- defN 16-Jan-01 00:00 wheel-0.38.4.dist-info/METADATA
?rw-r--r--  2.0 unx     2747 b- defN 16-Jan-01 00:00 wheel-0.38.4.dist-info/RECORD
```

Notice the file flag is `?` for all the generated files - `stat.S_IFREG` is not being set.

See https://github.com/pypa/wheel/issues/506, where this was just fixed in `wheel` (though it only had an issue with `RECORD`, while flit-core has this for more files). The fix was in https://github.com/pypa/wheel/pull/511. It should be fairly straightforward to apply the same fix here - `0o664 | stat.S_IFREG` should be used instead of 0o664 here: https://github.com/pypa/flit/blob/532bdaabafe0660b6c7494d8043ec4fa15b286a8/flit_core/flit_core/wheel.py#L140
Or it should be or'd in later, like when this function is called or inside this function: https://github.com/pypa/flit/blob/532bdaabafe0660b6c7494d8043ec4fa15b286a8/flit_core/flit_core/wheel.py#L35-L37
Actually, that function has a comment about making it a regular file by setting an extra bit but literally doesn't do it (and not sure what `0xFFFF` mention is, it's 0x8000, and and'ing with `0xFFFF` is done elsewhere). So that's where it probably should be done. But then again, that function runs on the actual regular files too, but those should be already regular files.

This doesn't "break" anything as far as I know, it's just slightly ugly with the `?`'s and technically incorrect.

henryiii:
Closes #634.

### DIFF
--- a/flit_core/flit_core/wheel.py
+++ b/flit_core/flit_core/wheel.py
@@ -33,7 +33,7 @@ def _write_wheel_file(f, supports_py2=False):
 
 
 def _set_zinfo_mode(zinfo, mode):
-    # Set the bits for the mode and bit 0xFFFF for “regular file”
+    # Set the bits for the mode
     zinfo.external_attr = mode << 16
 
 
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks @henryiii !

Issue #646, opened 2023-05-15
Missing runtime dependency on pip


stanislavlevin:
According to docs https://flit.pypa.io/en/stable/cmdline.html#flit-install:
> Flit calls pip to do the installation.

but `pip` is not specified as runtime dependency of `flit`.

takluyver:
Hmmm, good point. I think I just assume `pip` is always present, which naturally is not really true. Do you want to make a PR?

orsinium:
Project dependencies specify packages to be installed in the same environment as flit. However, flit currently uses pip installed in the target environment, which may be changed by passing `--python` flag. I'm going to change this behavior in #651 to make flit use pip from the flit's environment if available.

PR #647, opened 2023-05-16
deps: Add missing runtime dependency on pip

stanislavlevin:
According to docs https://flit.pypa.io/en/stable/cmdline.html#flit-install:
> Flit calls pip to do the installation.

but pip was not specified as runtime dependency of flit.

Fixes: https://github.com/pypa/flit/issues/646

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,6 +12,7 @@ dependencies = [
     "requests",
     "docutils",
     "tomli-w",
+    "pip",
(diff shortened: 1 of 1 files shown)


takluyver:
Closed and reopened because the CI didn't seem to have run for some reason. It's running now, and I think this change makes sense.

takluyver:
Thanks!

PR #649, opened 2023-06-20
Look for PyPI tokens in keyring when uploading

takluyver:
The username for keyring should look like `pypi_token:project:<project_name>` (using the [normalised](https://packaging.python.org/en/latest/specifications/name-normalization/) name), or `pypi_token:user:<username>`. If neither are found, it falls back to looking for a password (but if you have 2FA enabled on PyPI, uploads *have* to use tokens now).

I'd like to make some more tooling for conveniently creating tokens, but there isn't a PyPI API for it yet.

While I'm looking at this code, I've also simplified things a bit so that `FLIT_USERNAME` and `FLIT_PASSWORD` environment variables override details in `.pypirc` even if you use `--repository`, and it figures out the repository details before building the packages.

### DIFF
--- a/flit/upload.py
+++ b/flit/upload.py
@@ -9,11 +9,15 @@ import hashlib
 import logging
 import os
 from pathlib import Path
+import re
 import requests
 import sys
+from dataclasses import dataclass
+from typing import Optional
 from urllib.parse import urlparse
 
-from flit_core.common import Metadata
+from flit_core.common import make_metadata, Metadata, Module
+from .config import read_flit_config
 
 log = logging.getLogger(__name__)
 
@@ -27,6 +31,15 @@ SWITCH_TO_HTTPS = (
     "http://upload.pypi.io/",
 )
 
+
+@dataclass
+class RepoDetails:
+    url: str
+    username: Optional[str] = None
+    password: Optional[str] = None
+    is_warehouse: bool = True
+
+
 def get_repositories(file="~/.pypirc"):
     """Get the known repositories from a pypirc file.
 
@@ -38,9 +51,7 @@ def get_repositories(file="~/.pypirc"):
         file = os.path.expanduser(file)
 
         if not os.path.isfile(file):
-            return {'pypi': {
-                'url': PYPI, 'username': None, 'password': None,
-            }}
+            return {'pypi': RepoDetails(url=PYPI)}
 
         cp.read(file)
     else:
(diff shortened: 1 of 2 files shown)


takluyver:
I'm going to try this branch the next time I have a package to upload, before I merge it.

flit/upload.py
```diff
@@ -27,6 +31,15 @@
     "http://upload.pypi.io/",
 )
 
+
+@dataclass
+class RepoDetails:
+    url: str
+    username: Optional[str] = None
+    password: Optional[str] = None
+    is_warehouse: bool = True
```
Carreau, line 40: Do you want to update that naming to `is_legacy`?

flit/upload.py
```diff
@@ -160,21 +174,44 @@ def get_password(repo, prefer_env):
...
+    # https://packaging.python.org/en/latest/specifications/name-normalization/
+    project_name = re.sub(r"[-_.]+", "-", project_name).lower()
+    candidate_keys = [f"pypi_token:project:{project_name}"]
+
+    if repo.username is not None:
+        candidate_keys.append(f"pypi_token:user:{repo.username}")
+
+    try:
+        import keyring, keyring.errors
+    except ImportError:  # pragma: no cover
```
Carreau, line 203: Do you want to catch only `ModuleNotFoundError` and not all ImportErrors ?

Carreau (approved):
I should try that as well next time to make sure it works on MacOS

takluyver:
I forgot about this for a long while, and then for a while longer I didn't upload any packages through Flit. Now I just have again, and putting a token in my keyring with username `pypi_token:project:pep517` worked. :-)

flit/upload.py
takluyver: I think I prefer the broader `ImportError` - if importing keyring fails for any reason, treat it as not available. But in practice I don't think it's a big difference, as since [almost 3 years ago](https://github.com/jaraco/keyring/pull/480), keyring doesn't load its backends on import, so there's not much happening on import that's likely to fail.

flit/upload.py
takluyver: `is_pypi` might be a better name now - it's distinguishing PyPI (the new site, aka Warehouse) from other servers like devpi, and originally from PyPI classic.

It's only used for the 'Package is at' log message, so it's not too important that it's always right.

flit/upload.py
takluyver: I certainly tend to forget the difference, so the reminder is welcome. :slightly_smiling_face:

PR #651, opened 2023-07-11
Support environments without pip

orsinium:
Before: use pip installed in the target environment.

After: If pip is available in the current environment, use it, pointing it to the target environment using the `--python` flag. Otherwise, if Python is not installed in the same environment as flit, fall back to the old behavior.

This change is [inspired by rye](https://github.com/mitsuhiko/rye/blob/d920f8428c58edbf4552a90a0b413bfe4c2336d8/rye/src/installer.rs#L103). It allows to use flit with virtual environments created with `--without-pip` flag.

### DIFF
--- a/flit/install.py
+++ b/flit/install.py
@@ -249,8 +249,16 @@ class Installer(object):
             for req_d in requirements
         ]
 
+        # Use either pip in the venv or in the current environment,
+        # whichever is available.
+        try:
+            import pip  # noqa: F401
+        except ImportError:
+            cmd = [self.python, '-m', 'pip', 'install']
+        else:
+            cmd = [sys.executable, '-m', 'pip', 'install', '--python', self.python]
+
         # install the requirements with pip
-        cmd = [self.python, '-m', 'pip', 'install']
         if self.user:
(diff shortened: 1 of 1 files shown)


orsinium:
Right. I broke tests because they mock pip in the target environment. Well, tell me if you like the change, and I'll take care of actualizing tests.

takluyver:
Thanks, I think this makes sense, if people are going to rely on this new-ish pip feature to make virtualenvs without pip inside. Sorry it's taken me a while to look at it.

There's at least one more call to pip that should be changed in the same way:

https://github.com/pypa/flit/blob/6035005746ac1ee2f7952a9fd89431705ab0a5da/flit/install.py#L372

I think that's all, but please have a look yourself in case I've missed one.

Unfortunately I suspect this will be quite a pain for the tests, because as you spotted, we use the ability to point it to an arbitrary Python to mock out the installation command for the tests. If you get stuck with that, one way out might be to swap around the priority, e.g. doing `target_python -m pip --version` to check if it's installed in the target environment, and only using the `pip --python` option if not. :thinking:

PR #658, opened 2023-09-19
Remove destination data files if they exist.

antarcticrainforest:
We recently switched to flit and I came across this issue when I installed a package using the "symlink" mode. What I did was adding more data files and then re installing the package to create the symlinks to the recently added data files:

```console
flit install -s
FileExistsError: [Errno 17] File exists: '/home/wilfred/workspace/deployment/assets/share/freva/deployment/config/create_tables.sql' -> '/tmp/py_311/share/freva/deployment/config/create_tables.sql'
```

Since without `-s` any existing data files would be overridden anyways I figured it's ok to delete any existing data files.

### DIFF
--- a/flit/install.py
+++ b/flit/install.py
@@ -201,8 +201,7 @@ class Installer(object):
             rel_path = os.path.relpath(src_path, self.ini_info.data_directory)
             dst_path = os.path.join(target_data_dir, rel_path)
             os.makedirs(os.path.dirname(dst_path), exist_ok=True)
-            if os.path.exists(dst_path):
-                os.remove(dst_path)
+            pathlib.Path(dst_path, missing_ok=True)
             if self.symlink:
                 os.symlink(os.path.realpath(src_path), dst_path)
             else:
--- a/flit/install.py
+++ b/flit/install.py
@@ -201,7 +201,7 @@ class Installer(object):
             rel_path = os.path.relpath(src_path, self.ini_info.data_directory)
             dst_path = os.path.join(target_data_dir, rel_path)
             os.makedirs(os.path.dirname(dst_path), exist_ok=True)
-            pathlib.Path(dst_path, missing_ok=True)
+            pathlib.Path(dst_path).unlink(missing_ok=True)
             if self.symlink:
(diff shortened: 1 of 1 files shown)


takluyver:
Hi Martin! Long time no see. How are you?

Thanks, I think this change makes sense. Just for neatness, though, let's try deleting it and ignore the error if it's not there. Pathlib makes this easy with `.unlink(missing_ok=True)`. It won't usually make a difference, but it's one less way to hit odd behaviour if something else is adding/deleting files at the same time.

takluyver:
Thanks! I was actually just there last week - the next visit will probably be around early December, I'll let you know when I've figured it out.

Yeah, I'm wildly inconsistent on using pathlib vs os.path. :smile:

flit/install.py
```diff
@@ -201,6 +201,7 @@ def install_data_dir(self, target_data_dir):
             rel_path = os.path.relpath(src_path, self.ini_info.data_directory)
             dst_path = os.path.join(target_data_dir, rel_path)
             os.makedirs(os.path.dirname(dst_path), exist_ok=True)
+            pathlib.Path(dst_path, missing_ok=True)
```
takluyver, line 204:
```suggestion
            pathlib.Path(dst_path).unlink(missing_ok=True)
```

I think this is what we need.

There's a bug where `Path` currently accepts and ignores keyword arguments, so this line as is isn't an error. It looks like this is being fixed, with a warning from 3.12 and an error from 3.14. https://github.com/python/cpython/pull/19632

takluyver:
My bad, I had missed the note that `missing_ok` was added in Python 3.8.

I think we can drop Python 3.7 support from the main `flit` package now, as it's EOL anyway. I'll do a separate PR for that.

takluyver:
OK, that's done now in #660. I'll merge that tomorrow if no-one objects, then we can re-run the tests on this one.

takluyver:
Closing & reopening to re-run tests against the latest commit.

takluyver:
Tests are passing now. Thanks Martin!

PR #660, opened 2023-09-27
Require Python >=3.8 for flit CLI

takluyver:
Python 3.7 is already end of life (security updates ended earlier this year), so we can bump the minimum version up.

As usual, I'm being more conservative with the compatibility of `flit_core`, so you can still build packages on older versions of Python.

Prompted by #658.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -13,7 +13,7 @@ dependencies = [
     "docutils",
     "tomli-w",
 ]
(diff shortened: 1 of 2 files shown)


takluyver:
I'll leave this a day or so to give other people a chance to comment, but I don't expect it to be controversial - I seem to be more willing than most to keep supporting older Python versions.

PR #666, opened 2023-11-09
Documentation changes.

grandell1234:
In this request I fixed a few grammatical errors and added in another way to install Flit using `pip3 install flit` instead of `python3 -m pip install flit`.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -11,8 +11,9 @@ Install
 
     $ python3 -m pip install flit
 or
+
 ::
-    
+
     $ pip3 install flit
 
 Flit requires Python 3 and therefore needs to be installed using the Python 3
--- a/README.rst
+++ b/README.rst
@@ -11,6 +11,8 @@ Install
 
     $ python3 -m pip install flit
 or
+::
+    
     $ pip3 install flit
 
 Flit requires Python 3 and therefore needs to be installed using the Python 3
--- a/README.rst
+++ b/README.rst
@@ -1,5 +1,5 @@
 **Flit** is a simple way to put Python packages and modules on PyPI.
-It tries to require less thought about packaging and help you avoid common
+It tries to require less thought about packaging and helps you avoid common
 mistakes.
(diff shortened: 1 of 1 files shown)


Carreau (approved):

README.rst
```diff
@@ -1,5 +1,5 @@
 **Flit** is a simple way to put Python packages and modules on PyPI.
-It tries to require less thought about packaging and help you avoid common
+It tries to require less thought about packaging and helps you avoid common
```
joukewitteveen, line 2: "to require [...] and help" seems correct to me. "to require [...] and to help" is a bit redundant. The suggested "it tries to require [...] and [it] helps" can work, but I would not say it is better.

README.rst
```diff
@@ -53,7 +59,7 @@ or as a directory — and you want to distribute it.
        [project.urls]
        Home = "https://github.com/sirrobin/foobar"
 
-   You can edit this file to add other metadata, for example to set up
+   You can edit this file to add other metadata, for example, to set up
```
joukewitteveen, line 62: I wouldn't add this comma since the example is setting up command line scripts, not the addition of other metadata.

README.rst
```diff
@@ -68,7 +74,7 @@ or as a directory — and you want to distribute it.
 Once your package is published, people can install it using *pip* just like
 any other package. In most cases, pip will download a 'wheel' package, a
 standard format it knows how to install. If you specifically ask pip to install
-an 'sdist' package, it will install and use Flit in a temporary environment.
+a 'sdist' package, it will install and use Flit in a temporary environment.
```
joukewitteveen, line 77: You pronounce "sdist" starting with a consonant cluster? I would say S-dist, in which case "an" is the correct article. I don't know what the "official" pronunciation is.

joukewitteveen (changes requested):
Apart from mentioning `pip3 install flit`, I think these changes are no good.

PR #667, opened 2023-12-04
flit: improve the --python option

perillo:
pip --python option can take both the path to a Python executable or the path to a virtual environment, greatly improving the user experience.

Teach flit to do the same.

### DIFF
--- a/flit/__init__.py
+++ b/flit/__init__.py
@@ -26,6 +26,15 @@ def find_python_executable(python: Optional[str] = None) -> str:
         python = os.environ.get("FLIT_INSTALL_PYTHON")
     if not python:
         return sys.executable
+    if os.path.isdir(python):
+        # Assume it's a virtual environment and look for the environment's
+        # Python executable.  This is the same behavior used by pip.
+        #
+        # Try both Unix and Windows paths in case of odd cases like cygwin.
+        for exe in ("bin/python", "Scripts/python.exe"):
+            py = os.path.join(python, exe)
+            if os.path.exists(py):
+                python = py
     if os.path.isabs(python):  # sys.executable is absolute too
         return python
     # get absolute filepath of {python}
--- a/flit/__init__.py
+++ b/flit/__init__.py
@@ -34,7 +34,7 @@ def find_python_executable(python: Optional[str] = None) -> str:
         for exe in ("bin/python", "Scripts/python.exe"):
             py = os.path.join(python, exe)
             if os.path.exists(py):
-                python = py
+                return os.path.abspath(py)
     if os.path.isabs(python):  # sys.executable is absolute too
         return python
     # get absolute filepath of {python}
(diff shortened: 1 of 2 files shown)


perillo:
Hello, this is my first PR for this project.  I decided to adopt flit due to its simplicity.

The PR does not include a test, but it seems that there are no tests for the `--python` option.

flit/__init__.py
```diff
@@ -26,6 +26,15 @@ def find_python_executable(python: Optional[str] = None) -> str:
...
         return sys.executable
+    if os.path.isdir(python):
+        # Assume it's a virtual environment and look for the environment's
+        # Python executable.  This is the same behavior used by pip.
+        #
+        # Try both Unix and Windows paths in case of odd cases like cygwin.
+        for exe in ("bin/python", "Scripts/python.exe"):
+            py = os.path.join(python, exe)
+            if os.path.exists(py):
+                python = py
```
takluyver, line 37:
```suggestion
                return os.path.abspath(py)
```

If you pass e.g. `--python .venv`, we want to skip the code below which uses `shutil.which()`, I think.

takluyver:
Thanks for the contribution! I think this feature makes sense.

> The PR does not include a test, but it seems that there are no tests for the --python option.

Not for passing `--python` itself, but there are tests for the `find_python_executable()` function: https://github.com/pypa/flit/blob/main/tests/test_find_python_executable.py . Could you add a test for this? It's fine to do it in a simple way, e.g. using `/usr` and skipping the test if `/usr/bin/python` doesn't exist.

perillo:
> > The PR does not include a test, but it seems that there are no tests for the --python option.
> 
> Not for passing `--python` itself, but there are tests for the `find_python_executable()` function: https://github.com/pypa/flit/blob/main/tests/test_find_python_executable.py . Could you add a test for this? It's fine to do it in a simple way, e.g. using `/usr` and skipping the test if `/usr/bin/python` doesn't exist.

Ok.  I also noted that `codecov` complained about it.

What about a test like:
```python
def test_env(tmp_path):
    path = tmp_path / "venv"
    venv.create(path)
    executable = pathlib.Path(find_python_executable(path))
    assert executable.parent.parent.name == "venv" 
```
?

The test needs to import the `venv` and `pathlib` modules.

I would also like to add a test to ensure that a relative path to a venv is converted to an absolute path, but I'm not sure how to do it.

takluyver:
That sounds reasonable, thanks.

> I would also like to add a test to ensure that a relative path to a venv is converted to an absolute path, but I'm not sure how to do it. A possible solution is to use chdir but it may cause problems.

Use pytest's [monkeypatch fixture](https://docs.pytest.org/en/6.2.x/monkeypatch.html), and call `monkeypatch.chdir()`. This records that the CWD changed in the test, and pytest will change it back afterwards (whether the test passes or not).

perillo:
I avoided importing the `pathlib` module, since `find_python_executable` returns a string.  However I directly imported `isabs`, `basename` and `dirname` in order to reduce the line size in `test_env`.

takluyver:
Thanks @perillo, nice first contribution. :slightly_smiling_face:

perillo:
@takluyver Thanks to you.

PR #672, opened 2024-01-27
Update Sphinx & sphinx-rtd-theme for RTD

takluyver:
The docs build on #667 failed, complaining that some extension requires a newer version of Sphinx.


takluyver:
That seems to have worked.

Issue #675, opened 2024-02-27
Metadata 2.2+


groodt:
PyPI now supports Metadata 2.2+ and it would be ideal if flit started to build for newer standards by default

takluyver:
Thanks!

PR #676, opened 2024-03-11
Implement Metadata 2.3

groodt:
[PEP 643](https://peps.python.org/pep-0643/) introduces Metadata 2.2, where fields can either be static (default) or marked as dynamic. 
[PEP 685](https://peps.python.org/pep-0685/) introduces Metadata 2.3, that specifies how to normalize extra names.

This PR bumps the Metadata version to 2.3. `flit` has always written and published static metadata in both wheel `METADATA` and sdist `PKG-INFO` and therefore this PR does not intend to add new support for recording dynamic metadata into distributions. This should not be confused with the `dynamic` metadata in the pyproject.toml files, which are dynamic in flit and remain supported (PEP 621).

Upgrading to the newer Core Metadata standards has advantages for the python packaging ecosystem. It makes it possible for resolvers to read metadata directly without a PEP 517 invocation, which can bring large speedups.

Towards: https://github.com/pypa/flit/issues/675

### DIFF
--- a/flit_core/flit_core/common.py
+++ b/flit_core/flit_core/common.py
@@ -366,6 +366,24 @@ class Metadata(object):
         # Normalized Names (PEP 503)
         return re.sub(r"[-_.]+", "-", name).lower()
 
+    def _extract_extras(self, req):
+        match = re.search(r'\[([^]]*)\]', req)
+        if match:
+            list_str = match.group(1)
+            return [item.strip() for item in list_str.split(',')]
+        else:
+            return None
+    
+    def _normalise_requires_dist(self, req):
+        extras = self._extract_extras(req)
+        if extras:
+            normalised_extras = [self._normalise_core_metadata_name(extra) for extra in extras]
+            normalised_extras_str = ', '.join(normalised_extras)
+            normalised_req = re.sub(r'\[([^]]*)\]', f"[{normalised_extras_str}]", req)
+            return normalised_req
+        else:
+            return req
+
     def write_metadata_file(self, fp):
         """Write out metadata in the email headers format"""
         fields = [
@@ -404,7 +422,7 @@ class Metadata(object):
             fp.write(u'Classifier: {}\n'.format(clsfr))
 
         for req in self.requires_dist:
-            normalised_req = self._normalise_core_metadata_name(req)
+            normalised_req = self._normalise_requires_dist(req)
             fp.write(u'Requires-Dist: {}\n'.format(normalised_req))
 
         for url in self.project_urls:
--- a/flit_core/flit_core/common.py
+++ b/flit_core/flit_core/common.py
@@ -347,6 +347,7 @@ class Metadata(object):
     obsoletes_dist = ()
     requires_external = ()
     provides_extra = ()
+    dynamic = ()
 
     metadata_version = "2.3"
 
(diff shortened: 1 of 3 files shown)


groodt:
Thinking about this more, I think I’m definitely over normalizing Requires-Dist if there are any markers. I’ll find some time to change it so that it only normalizes any extras.

groodt:
I've added some test cases. I believe that this is now ready for review.

flit_core/flit_core/common.py
```diff
@@ -348,7 +348,7 @@ class Metadata(object):
     requires_external = ()
     provides_extra = ()
```
takluyver, line 349:
```suggestion
    provides_extra = ()
    dynamic = ()
```

Let's add the field for completeness even if we're not using it. Also in the `write_metadata_file()` method, so we don't overlook it if there ever is a need for it.

takluyver:
Thanks, this is looking good to me. :slightly_smiling_face: 

One minor wrinkle is that if we fall back to executing the module to get version & description, we can't actually guarantee that they won't have a different value if you build a wheel from the sdist. You could theoretically construct them so they differ based on the day of the week, or the system it's built on. But the spec also says version can't be dynamic, and hopefully no-one actually does that, so let's cross our fingers and just say it's static.

I wonder if it would make sense to normalise values on the way in to `Metadata`, so checking e.g. `md.provides_extra == other.provides_extra` uses the normalised values. But I don't think we do any comparisons like that anywhere in the code, so let's leave it as you've done it unless we see a concrete reason to change.

takluyver:
Thanks @groodt !

groodt:
My pleasure!

Although, I think this might have been merged prematurely. I should have
moved it to draft.

I have doubts over how I currently implemented it. See
https://discuss.python.org/t/status-or-funding-for-metadata-2-2-in-pypi/46137/36

takluyver:
Ah, good point. Sorry, I didn't notice your message until now.

My reading of both PEP 685 and the [core metadata spec](https://packaging.python.org/en/latest/specifications/core-metadata/#requires-dist-multiple-use) is that they only explicitly say the `[extra,names]` you require of another project have to be normalised. But I think the intent is that optional dependencies with an `extra='foo'` environment marker should probably be normalised too, and it seems best to do so. Would you be interested in making a follow-up PR to do that?

Looking at PEP 685, we should also make sure we throw suitable errors if invalid or clashing extras are specified.

groodt:
@takluyver Yes, I'm happy to pick it up as a follow-up. Are we sure that I'm not breaking anything with the currently merged code?

I guess I'm not familiar enough with the 2 styles of extras from a library author point of view. I use extras often when I'm building apps.

Are these both valid `Requires-Dist` entries? It's unclear from the PEPs I've read.

* `Requires-Dist: foo[bar]`
* `Requires-Dist: foo; extra == 'bar'

Are they semantically different?

merwok:
requires `foo[bar]` means: I want foo, with its extra bar (and the dependencies for both)

requires `foo; extra == bar` means: I want foo, if I am installed with my extra bar

groodt:
Thanks @merwok That makes sense!

takluyver:
I'd be surprised if things break with the current state - I'd guess pip & other package consumer tools will normalise extras in environment markers (`extra == 'bar'`) anyway, because even if they read the spec differently, they need to handle packages with older metadata versions, and it's easier to normalise in all cases than check the metadata version first.

But to be on the safe side, I'd ideally like to normalise those as well before releasing `flit_core` claiming to produce metadata 2.3. I don't see any downside to making extras in environment markers consistent with `Provides-Extra` fields.

PR #682, opened 2024-05-14
datetime.utcfromtimestamp is deprecated in Python 3.12, avoid using it

wimglenn:
https://docs.python.org/3/library/datetime.html#datetime.datetime.utcfromtimestamp

The alternative proposed should work in all Python versions flit supports

### DIFF
--- a/flit_core/flit_core/wheel.py
+++ b/flit_core/flit_core/wheel.py
@@ -2,6 +2,7 @@ import argparse
 from base64 import urlsafe_b64encode
 import contextlib
 from datetime import datetime
+from datetime import timezone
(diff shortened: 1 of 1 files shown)


flit_core/flit_core/wheel.py
```diff
@@ -2,6 +2,7 @@
 from base64 import urlsafe_b64encode
 import contextlib
 from datetime import datetime
+from datetime import timezone
```
takluyver, line 5:
```suggestion
from datetime import datetime, timezone
```

takluyver:
Thanks!

PR #684, opened 2024-05-28
MNT: fix compatibility with Python 3.14

tacaswell:
The ast.Str class was deprecated in 3.8 and will be removed in 3.14

### DIFF
--- a/flit_core/flit_core/common.py
+++ b/flit_core/flit_core/common.py
@@ -148,6 +148,10 @@ def get_docstring_and_version_via_ast(target):
         with target_path.open('rb') as f:
             node = ast.parse(f.read())
         for child in node.body:
+            if sys.version_info >= (3, 8):
+                target_type = ast.Constant
+            else:
+                target_type = ast.Str
             # Only use the version from the given module if it's a simple
             # string assignment to __version__
             is_version_str = (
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks, nice catch! `flit_core` still supports Python down to 3.6, so could I ask you to do a little bit of compatibility code here? Something simple like `if sys.version_info >= (3, 8)` should be fine.

I know I'm pretty weirdly conservative on this. In most projects I'd have dropped such old Python versions some time ago. But `flit_core` is kind of in a special position, being used to build a whole lot of other packages, including key packaging tools like `build` & `installer`. [Python 3.6 is still accounting for 2-3% of downloads from PyPI](https://pypistats.org/packages/__all__), so I'd rather keep it compatible than have some users getting an older version of flit_core.

takluyver:
Thanks Tom!

AA-Turner:
Sorry to be one of those people, but would it be possible to have a release with this commit? I'm currently unable to test with Python 3.14 in CI as ``pip`` is failing with this error.

A

pmp-p:
Sorry to be one of those people **too**, but please would it be possible to have a release with this commit? I'm currently unable to test with Python 3.14 in CI as pip is failing with this error.

please.

AA-Turner:
Thanks @takluyver! We run Sphinx on the tip of CPython for doctests so testing the latest release is useful, but I agree perhaps less so in the general case.

A

PR #689, opened 2024-08-19
Add an exclude patttern for wheels (and use it for flit-core)

cjmayo:
Allows files to be included in the sdist but not installed.
Useful for established projects that have tests as a subdirectory of the
package directory.

----

flit-core 3.9.0 installs an `__init__.py` and 6 submodules plus 161 files (including byte-code files) in 44 directories
under a tests directory.

Draft because there is no documentation or tests at the moment. If there is a likelihood of it getting merged I can have a go at those.

### DIFF
--- a/flit_core/flit_core/common.py
+++ b/flit_core/flit_core/common.py
@@ -1,8 +1,10 @@
 import ast
 from contextlib import contextmanager
+from glob import glob
 import hashlib
 import logging
 import os
+import os.path as osp
 import sys
 
 from pathlib import Path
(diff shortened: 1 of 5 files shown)


takluyver:
I've opened #691, which moves the `flit_core` tests out of the package folder, as a point of comparison for this.

takluyver:
It certainly makes sense to include tests in the sdist, but I don't see anything in the Gentoo info saying that the tests need to be inside the package directory - and I'd be surprised if Gentoo expected that, given that major Python packages like [requests](https://github.com/psf/requests) and [pip](https://github.com/pypa/pip) have their tests outside the package directory.

I'm not too concerned about the implementation complexity - though that is a small factor - but in the long run I think it's better for everyone to have fewer different ways to structure a project. We're not going to get rid of the choice between tests inside the package folder & installed vs. outside & not installed, but I don't see a good argument for tests inside the package folder but *not* installed. I accept that a project already set up that way & migrating to Flit would have an easier time in the short term if that was an option. But longer term I think it's better to push them to pick one of the 2 options that all packages built with Flit use, rather than creating a third option so they don't have to change.

I'm just about to go away for a few days. I'm thinking that if no more compelling use case has come up by the time I get back, then I'll close this and merge #691. Of course, if you think I'm not getting something or not giving something sufficient weight, feel free to make that case too. And thanks for the idea & the discussion, in any case. :slightly_smiling_face:

takluyver:
Thanks for the discussion - for the reasons given I'm going to close this and merge #691.

PR #695, opened 2024-10-18
Prepare release notes for Flit 3.10

takluyver:


takluyver:
(Note to self) I wanted to do something to follow up on #676 before making a new release.

takluyver:
I want to land #697 to ensure we're complying properly with the metadata 2.3 changes ([PEP 685](https://peps.python.org/pep-0685/)).

PR #699, opened 2024-10-28
Pypi.io does not exists anymore.

Carreau:
It was a temporary url, still redirect for some GET but should not be used.


takluyver:
Thanks :-)

PR #701, opened 2024-10-31
Provide token to upload coverage data to codecov

takluyver:
Codecov is adding an extra ~20 seconds to each CI job, because it hits a rate limit and retries several times before timing out. I think giving it the token should work around this.


takluyver:
Yup, that seems to be working much better. :tada:

Issue #692, opened 2024-09-03
[Feature request] Implement PEP 639


befeleme:
PEP 639 has been provisionally accepted with one of the conditions being implementation in two build backends.
See: https://peps.python.org/pep-0639

Additional context:
- The full implementation, including the Core Metadata bump to 2.4, will require also the support on the side of PyPI first: https://github.com/pypi/warehouse/issues/16620
- There's a plan to add validation data and parser to packaging: https://github.com/pypa/packaging/pull/799

cdce8p:
I've opened PRs to add support for both `License-File` and `License-Expression`, see #705 and #707. The final support for version `2.4` is still dependent on upstream support. AFAIK `twine` still doesn't support it.

cdce8p:
@takluyver Did you had a chance to look at #705 + #707 yet? I'd love to move forward with those and get support for PEP 639 implemented.

AA-Turner:
This has now been done (thanks to @cdce8p!).

A

PR #705, opened 2024-11-20
Add initial support for license-files

cdce8p:
Add support for the `license-files` key in `pyproject.toml`.
https://peps.python.org/pep-0639/#add-license-files-key
https://peps.python.org/pep-0639/#add-license-file-field

This PR only adds the required validation and ensures the license files are properly packaged in the `.dist-info/licenses` folder. It doesn't add the `License-File` key to the metadata as that requires full support for metadata version `2.4`.

Work on #692


cdce8p:
@takluyver Would love to get your opinion here. I do plan to continue working on the PEP 639 support afterwards.

flit_core/flit_core/config.py
```diff
@@ -73,6 +74,8 @@ class ConfigError(ValueError):
     'dynamic',
 }
 
+default_license_files_globs = ['COPYING*', 'LICENSE*']
```
AA-Turner, line 77: Would you consider supporting "Licence*"? (The non-US English spelling) 

[Setuptools includes](https://setuptools.pypa.io/en/latest/userguide/miscellaneous.html) COPYING, LICENCE, LICENSE, NOTICE, and AUTHORS by default, for context -- I'm not proposing to include the latter two here.

```suggestion
default_license_files_globs = ['COPYING*', 'LICEN[CS]E*']
```

cdce8p: Would be fine with me. I just copied the existing values. Seems like they were added in https://github.com/pypa/flit/pull/99#discussion_r97009036

doc/pyproject_toml.rst
```diff
@@ -98,6 +98,8 @@ requires-python
 license
   A table with either a ``file`` key (a relative path to a license file) or a
   ``text`` key (the license text).
+license-files
+  A list of glob patterns for license files to include. Defaults to ``['COPYING*', 'LICENSE*']``.
```
AA-Turner, line 102: To match the other change
```suggestion
  A list of glob patterns for license files to include. Defaults to ``['COPYING*', 'LICEN[CS]E*']``.
```

flit_core/flit_core/config.py
AA-Turner: Thanks -- I've added another GH suggestion for the other place that would need to change.

flit_core/flit_core/config.py
```diff
@@ -439,6 +449,37 @@ def _expand_requires_extra(re):
...
+                "Invalid glob pattern for [project.license-files]: '{}'. "
+                "Pattern must not contain '..'".format(pattern)
+            )
+        try:
+            files = [
+                str(file.relative_to(project_dir)).replace(osp.sep, "/")
+                for file in project_dir.glob(pattern)
+                if file.is_file()
+            ]
+        except ValueError as ex:
```
takluyver, line 471: There's another check we should do on the patterns.

> Any characters or character sequences not covered by this specification are invalid. Projects MUST NOT use such values. Tools consuming this field SHOULD reject invalid values with an error.

The spec mentions alphanumerics (for which we can rely on `str.isalnum()`), plus `_-.`, globbing characters `*?[]` and the directory separator `/`. Everything else should be rejected.

flit_core/flit_core/config.py
```diff
@@ -551,6 +592,18 @@ def read_pep621_metadata(proj, path) -> LoadedConfig:
                 "file or text field required in [project.license] table"
             )
 
+    license_files = []
+    if 'license-files' in proj:
+        _check_type(proj, 'license-files', list)
+        globs = proj['license-files']
+        license_files = _license_files_from_globs(path.parent, globs)
```
takluyver, line 599: We should prevent mixing the new-style `license-files` definition with the older `license` table.

> If the new license-files key is present, build tools MUST raise an error if the license key is defined and has a value other than a single top-level string.

```suggestion
    if 'license-files' in proj:
        _check_type(proj, 'license-files', list)
        globs = proj['license-files']
        license_files = _license_files_from_globs(path.parent, globs)
        if isinstance(proj.get('license', None), dict):
            raise ConfigError("license-files replaces license table, do not use both")
```

(The spec also says tools SHOULD add deprecation warnings, but I like to enable the new thing before I start prodding people to move off the old, so let's leave that for after doing a release)

flit_core/flit_core/config.py
```diff
@@ -551,6 +592,18 @@ def read_pep621_metadata(proj, path) -> LoadedConfig:
...
+    else:
+        license_files = _license_files_from_globs(
+            path.parent, default_license_files_globs, use_defaults=True
+        )
```
takluyver, line 603: If `project.license.file` is specified, we can emit that as `License-File` metadata.

> If the specified license `file` is present in the source tree, build tools SHOULD use it to fill the License-File field in the core metadata, and MUST include the specified file as if it were specified in a license-file field.

flit_core/tests_core/test_config.py
```diff
@@ -178,3 +198,28 @@ def test_bad_pep621_readme(readme, err_match):
...
+    path = samples_dir / 'pep621_license_files' / 'pyproject.toml'
+    data = path.read_text()
+    data = re.sub(
+        r"(^license-files = )(?:\[.*\])", r"\g<1>{}".format(value),
+        data, count=1, flags=re.M
+    )
+    dir = os.getcwd()
+    try:
+        os.chdir(samples_dir / 'pep621_license_files')
+        with patch("pathlib.Path.read_text", return_value=data):
```
takluyver, line 221: Can I ask that we test with `read_pep621_metadata()` or `prep_toml_config()`, which both take dicts as input, rather than mocking the file read and regexing replacement data into place? I made a similar change in #712 (6d60848cea0d64a4ec69a55441bf717da9e7fcfc) if it's not clear what I mean. We don't need to test the TOML parser here.

flit_core/tests_core/test_wheel.py
```diff
@@ -45,3 +46,16 @@ def test_data_dir(tmp_path):
     assert_isfile(info.file)
     with ZipFile(info.file, 'r') as zf:
         assert 'module1-0.1.data/data/share/man/man1/foo.1' in zf.namelist()
+
+
+def test_license_files(tmp_path):
+    dir = os.getcwd()
+    try:
+        os.chdir(samples_dir / 'pep621_license_files')
```
takluyver, line 54: I don't think we need to modify the CWD here - e.g. the test just above doesn't do this.

takluyver (commented):
Thanks, I think this basically looks good, but there are a few extra checks that I think should occur on the config.

takluyver:
Thanks!

PR #715, opened 2025-02-08
Package metadata version 2.4

takluyver:
Enable the new License-Expression and License-Files fields.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,9 +16,9 @@ dependencies = [
 ]
 requires-python = ">=3.8"
 readme = "README.rst"
-license = {file = "LICENSE"}
+license = "BSD-3-Clause"
+license-files = ["LICENSE"]
 classifiers = ["Intended Audience :: Developers",
-    "License :: OSI Approved :: BSD License",
     "Programming Language :: Python :: 3",
     "Topic :: Software Development :: Libraries :: Python Modules",
 ]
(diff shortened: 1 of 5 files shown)


cdce8p (approved):
🎉

pyproject.toml
```diff
@@ -16,9 +16,9 @@ dependencies = [
 ]
 requires-python = ">=3.8"
 readme = "README.rst"
-license = {file = "LICENSE"}
+license = "BSD-3-Clause"
+license-files = ["LICENSE"]
```
cdce8p, line 20: Should the `build-system.requires` be bumped for this one to `>=3.11`?
https://github.com/pypa/flit/blob/2818f41fb2a6aedfd264c175ed3ca1434f39674b/pyproject.toml#L1-L3

takluyver: Good point, yes!

takluyver:
I'll give this a couple of days for other contributors to take a look, if no problems come up I'll merge it and make a new release.

cdce8p:
Opened #716 and #717 with small improvements for the license handling and docs. Would make sense to include these in 3.11 as well, I believe.

cdce8p:
> I'll give this a couple of days for other contributors to take a look, if no problems come up I'll merge it and make a new release.

I tested it with some other packages. Overall it looks great. The main issue I encountered was with validation, in particular complex license expressions. Opened #718 for it.

If you're open to it, I don't expect any issues with existing configs. So it would be possible to do the release now as well and allow end users to test it. Potential new issues, if there are any, could then be fixed in a `3.11.1` release shortly there after.

takluyver:
I spotted one more bit to update for the license metadata changes, which is the basic project setup created by `flit init`. I've done that in #719.

PR #712, opened 2025-01-20
Support simple SPDX license expressions (PEP 639)

takluyver:
This follows on from #707. I plan to allow simple license expressions initially - referring to one license and no exceptions - but validate & normalise them properly. Compound license expressions can be added later.

PR #716, opened 2025-02-08
Add support for license.text

cdce8p:
Before license expressions, the license table was allowed to contain a `text` key. That never got implemented even though the docs list it as supported. Map `license.text` to the `License` core metadata field.
https://github.com/pypa/flit/blob/2818f41fb2a6aedfd264c175ed3ca1434f39674b/doc/pyproject_toml.rst?plain=1#L98-L101

It seems this was accidentally removed in https://github.com/pypa/flit/pull/712/commits/f08bb9f0c6dee63b38f0d1e022ffbd8c67fb683b

### DIFF
--- a/flit_core/flit_core/config.py
+++ b/flit_core/flit_core/config.py
@@ -612,7 +612,9 @@ def read_pep621_metadata(proj, path) -> LoadedConfig:
                     raise ConfigError(f"License file {license_f} does not exist")
                 license_files.add(license_tbl['file'])
             elif 'text' in license_tbl:
-                pass
+                license = license_tbl['text']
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks, but I actually took it back out deliberately. It was never clear to me whether the `project.license.text` field was meant to contain a license name or the full text of the license, and I didn't want to mess up the metadata with the full text, especially since it wasn't really used for anything. Before the license expressions, the classifiers were the best way to add vaguely structured info about licenses. And it seems a bit weird to start adding that in just when it's deprecated.

Sorry not to be clearer about that and leave you redoing the change!

PR #718, opened 2025-02-08
Don't fail for complex license expressions

cdce8p:
Flit should support complex license expressions with `AND`, `OR` or `WITH` and not raise `ConfigError`.
We can skip the validation here (for now). PyPI will always validated uploads.

### DIFF
--- a/flit_core/flit_core/config.py
+++ b/flit_core/flit_core/config.py
@@ -823,6 +823,10 @@ def normalize_license_expr(s: str):
     """
     from ._spdx_data import licenses
     ls = s.lower()
+    if any(keyword in ls for keyword in ("and", "or", "with")):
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks, but my preference is to allow only simple license expressions for now, and broaden it to valid compound expressions in the future, so we're expanding the set of what Flit accepts, not narrowing it (as discussed [in #707](https://github.com/pypa/flit/pull/707#issuecomment-2572908404)). PyPI will validate license expressions on upload, but private indexes may not, and people may also just build packages without uploading them anywhere.

If you really want to use compound expressions in the meantime,  you can set the environment variable `FLIT_ALLOW_INVALID=1` to explicitly bypass the check.

PR #725, opened 2025-02-20
Add additional check for project.license.file

### LINKED ISSUE #724: `license.file` failure in flit-core 3.11
Consider an exemplar project, `example`, in a directory containing only a blank `example.py` and the following `pyproject.toml`:

```toml
[build-system]
requires = ["flit_core~=3.10.0"]

[project]
name = "example"
version = "1"
description = "example"
license.file = "LICENCE.rst"
requires-python = ">=3.9"
```

Note that the licence file, `LICENCE.rst`, does not exist in the project directory. With *flit-core* 3.10, wheels build with `python -m build --wheel`, and the project is installable with `python -m pip install .`. Sdist creation fails, but this is not required for `pip install .`.

Updating to *flit-core* 3.11, we observe an error noting that the licence file does not exist (`flit_core.config.ConfigError: License file LICENCE.rst does not exist`). Suppose that the licence file is actually located in the parent directory of the `pyproject.toml` file (e.g. a multi-language project, or a mono-repo). An initial attempt at using relative paths (`licence.file = "../LICENCE.rst"` finishes with no errors, but the file is not included in the sdist, and is included directly in `.dist-info` in the wheel, rather than in the `licences` directory.

Thus my questions:

1. Is it possible to preserve *flit-core* 3.10 behaviour when building wheels if the licence file given in `licence.file` does not exist? I'm unsure if this is compatible with Metadata 2.4, though.
2. Are relative paths in `license.file` (e.g. `../LICENCE.rst`) allowed? [1]
3. If (2), is there a bug with relative file handling in source/built distributions?
4. If not (2), can *flit-core* raise an error?

Thanks,
Adam

cc @cdce8p 

[1] [PEP 639 forbids](https://peps.python.org/pep-0639/#add-license-files-key) such paths in `licence-files`, but I'm unclear on `licence.file`. [PEP 621 is silent](https://peps.python.org/pep-0621/#license) on the matter, and unhelpfully the `text` and `file` tables seem to have been removed from [the specification](https://packaging.python.org/en/latest/specifications/pyproject-toml/#license), even though they're only deprecated.

----

P.S. If useful, the project where this came up is https://github.com/AA-Turner/Roman-Numerals. I've tried to minimise the above and make it context-free, though. I'm prevented from updating to the PEP 639 style metadata in this project as the licence expression is `0BSD OR CC-1.0`. This is just the first time that a new Flit release has ever caused an error for me, though -- a testament to the project.

cdce8p:
Closes #724

### DIFF
--- a/flit_core/flit_core/config.py
+++ b/flit_core/flit_core/config.py
@@ -473,7 +473,7 @@ def _license_files_from_globs(project_dir: Path, globs, warn_no_files = True):
             )
         try:
             files = [
-                str(file.relative_to(project_dir)).replace(osp.sep, "/")
+                file.relative_to(project_dir).as_posix()
                 for file in project_dir.glob(pattern)
                 if file.is_file()
             ]
@@ -615,7 +615,7 @@ def read_pep621_metadata(proj, path) -> LoadedConfig:
                 license_p = path.parent / license_f
                 if not license_p.is_file():
                     raise ConfigError(f"License file {license_tbl['file']} does not exist")
-                license_f = str(license_p.relative_to(path.parent)).replace(osp.sep, "/")
+                license_f = license_p.relative_to(path.parent).as_posix()
                 license_files.add(license_f)
             elif 'text' in license_tbl:
                 pass
--- a/flit_core/flit_core/config.py
+++ b/flit_core/flit_core/config.py
@@ -603,17 +603,19 @@ def read_pep621_metadata(proj, path) -> LoadedConfig:
                     raise ConfigError(
                         "[project.license] should specify file or text, not both"
                     )
-                license_f = license_tbl['file']
+                license_f = osp.normpath(license_tbl['file'])
                 if isabs_ish(license_f):
                     raise ConfigError(
-                        f"License file path ({license_f}) cannot be an absolute path"
+                        f"License file path ({license_tbl['file']}) cannot be an absolute path"
                     )
-                if ".." in license_f:
+                if license_f.startswith('..' + os.sep):
                     raise ConfigError(
-                        f"License file path ({license_f}) cannot contain '..'"
+                        f"License file path ({license_tbl['file']}) cannot contain '..'"
                     )
-                if not (path.parent / license_f).is_file():
-                    raise ConfigError(f"License file {license_f} does not exist")
+                license_p = path.parent / license_f
+                if not license_p.is_file():
+                    raise ConfigError(f"License file {license_tbl['file']} does not exist")
+                license_f = str(license_p.relative_to(path.parent)).replace(osp.sep, "/")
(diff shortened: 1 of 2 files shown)


flit_core/flit_core/config.py
```diff
@@ -608,9 +608,13 @@ def read_pep621_metadata(proj, path) -> LoadedConfig:
                     raise ConfigError(
                         f"License file path ({license_f}) cannot be an absolute path"
                     )
+                if ".." in license_f:
```
AA-Turner, line 611: Perhaps? Although, `normpath` just does lexical manipulation and e.g. `a/spam/../b` might refer to a symlink.

```suggestion
                if ".." in os.path.normpath(license_f):
```

AA-Turner: Or actually, as e.g. a file might be named `a..b`. I think backslashes are prohibited?


```suggestion
                if ".." in license_f.split('/'):
```

cdce8p: > Perhaps? Although, `normpath` just does lexical manipulation and e.g. `a/spam/../b` might refer to a symlink.

Technically, yes. Although I'm not sure it's worth the effort. Same for file names which contain `..`.
For the overwhelming majority, checking `if ".." in license_f:` to provide a better error message should be enough.

takluyver: Here's the check we do for paths in the sdist include/exclude lists:

https://github.com/pypa/flit/blob/5f8c75fba2f926036da6188ab8b59524ea4859f3/flit_core/flit_core/config.py#L239-L249

I can't think of a use case for a path like `foo/../bar` (which will normalise to just `bar`), but it is currently allowed in that case. Whatever we do, I'd like it to be consistent across all the places where a relative path is specified.

flit_core/flit_core/config.py
```diff
@@ -603,14 +603,20 @@ def read_pep621_metadata(proj, path) -> LoadedConfig:
...
-                    raise ConfigError(f"License file {license_f} does not exist")
-                license_files.add(license_tbl['file'])
+                if license_f.startswith('..' + os.sep):
+                    raise ConfigError(
+                        f"License file path ({license_tbl['file']}) cannot contain '..'"
+                    )
+                license_p = path.parent / license_f
+                if not license_p.is_file():
+                    raise ConfigError(f"License file {license_tbl['file']} does not exist")
+                license_f = str(license_p.relative_to(path.parent)).replace(osp.sep, "/")
```
AA-Turner, line 618:
```suggestion
                license_f = license_p.relative_to(path.parent).as_posix()
```

AA-Turner (approved):

takluyver:
Thanks!

PR #728, opened 2025-02-23
Support annotated assignment for ```__version__``` strings

AA-Turner:
I recently tried to update a package to use a ``Final`` annotation for the ``__version__`` attribute, but realised that Flit doesn't support annotated assignment. This PR adds support and tests.

A

### DIFF
--- a/flit_core/flit_core/common.py
+++ b/flit_core/flit_core/common.py
@@ -148,28 +148,27 @@ def get_docstring_and_version_via_ast(target):
         with target_path.open('rb') as f:
             node = ast.parse(f.read())
         for child in node.body:
-            if sys.version_info >= (3, 8):
-                target_type = ast.Constant
-            else:
-                target_type = ast.Str
-            # Only use the version from the given module if it's a simple
-            # string assignment to __version__
-            is_version_str = (
-                    isinstance(child, ast.Assign)
-                    and any(
-                        isinstance(target, ast.Name)
-                        and target.id == "__version__"
-                        for target in child.targets
-                    )
-                    and isinstance(child.value, target_type)
-            )
-            if is_version_str:
+            if is_version_str_node(child):
                 if sys.version_info >= (3, 8):
                     version = child.value.value
                 else:
                     version = child.value.s
                 break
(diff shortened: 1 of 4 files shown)


flit_core/flit_core/common.py
```diff
@@ -172,6 +157,20 @@ def get_docstring_and_version_via_ast(target):
     return ast.get_docstring(node), version
 
 
+def is_version_str_node(node, /):
```
takluyver, line 160:
```suggestion
def is_version_str_node(node):
```

For compatibility with older Python versions.

We're probably coming to the point where even for flit_core I'll drop support for some older Python versions, but let's work that out separately. I generally find it easier to support a pretty broad range of Python versions than to deal with people getting stuck on an old version of a package I maintain because I dropped support for older Python.

flit_core/flit_core/common.py
```diff
@@ -148,22 +148,7 @@ def get_docstring_and_version_via_ast(target):
...
-                    isinstance(child, ast.Assign)
-                    and any(
-                        isinstance(target, ast.Name)
-                        and target.id == "__version__"
-                        for target in child.targets
-                    )
-                    and isinstance(child.value, target_type)
-            )
-            if is_version_str:
+            if is_version_str_node(child):
```
AA-Turner, line 151:
```suggestion
            if is_version_str_assignment(child):
```

flit_core/flit_core/common.py
AA-Turner, line 160:
```suggestion
def is_version_str_assignment(node):
```

AA-Turner: Done

henryiii: The convention for mypy is that a name starting with `__*` is positional-only. So as far as mypy is concerned, `...(__node):` is identical to `...(node, /)`. (If you really wanted it to be positional only).

AA-Turner: Indeed, but it doesn't matter that much (all of this is internal-only, anyway).

takluyver:
Thanks!

PR #731, opened 2025-03-01
Support compound SPDX licence expressions

### LINKED ISSUE #723: Support for SPDX composite license expressions
To get PEP 639 license expressions in without too much delay, I opted to validate only simple expressions, i.e. a single license. SPDX also allows [composite license expressions](https://spdx.github.io/spdx-spec/v2.2.2/SPDX-license-expressions/#d4-composite-license-expressions), combining multiple licenses with AND & OR, and detailing exceptions with WITH. Validation should be extended to allow these.

I'd like to aim to do this without any external dependencies, keeping flit_core with zero dependencies on recent Python versions, even though that will mean some duplicated effort with other packaging related projects.

AA-Turner:
Closes #723.

cc @takluyver @cdce8p 

A

### DIFF
--- a/flit_core/flit_core/config.py
+++ b/flit_core/flit_core/config.py
@@ -840,18 +840,18 @@ def normalise_compound_license_expr(s: str) -> str:
     parts = []
     try:
         for part in filter(None, re.split(r' +|([()])', s)):
-            if part in {'WITH', 'with'}:
+            if part.upper() == 'WITH':
                 # provide a sensible error message for the WITH operator
                 raise ConfigError(f"The SPDX 'WITH' operator is not yet supported!")
-            elif part in {'and', 'or', 'with'}:
-                # provide a sensible error message for lowercase operators
-                reason = f"operators must be uppercase, not '{s}'"
-                raise ConfigError(invalid_msg.format(s=s, reason=reason))
             elif part in {'AND', 'OR'}:
                 if not parts or parts[-1] in {' AND ', ' OR ', ' WITH ', '('}:
                     reason = f"'{part}' must follow a license ID"
                     raise ConfigError(invalid_msg.format(s=s, reason=reason))
                 parts.append(f' {part} ')
+            elif part.lower() in {'and', 'or', 'with'}:
+                # provide a sensible error message for non-uppercase operators
+                reason = f"operators must be uppercase, not '{s}'"
+                raise ConfigError(invalid_msg.format(s=s, reason=reason))
             elif part == '(':
                 if parts and parts[-1] not in {' AND ', ' OR ', '('}:
                     reason = "'(' must follow a license expression operator or another '('"
--- a/flit_core/flit_core/config.py
+++ b/flit_core/flit_core/config.py
@@ -824,7 +824,7 @@ def isabs_ish(path):
     return os.path.isabs(path) or path.startswith(('/', '\\'))
 
 
-def normalise_compound_license_expr(s: str):
+def normalise_compound_license_expr(s: str) -> str:
     """Validate and normalise a compund SPDX license expression.
 
     Per the specification, licence expression operators (AND, OR and WITH)
@@ -872,7 +872,7 @@ def normalise_compound_license_expr(s: str):
     return ''.join(parts)
 
 
-def normalise_simple_license_expr(s: str):
+def normalise_simple_license_expr(s: str) -> str:
     """Normalise a simple SPDX license expression.
 
     https://spdx.github.io/spdx-spec/v2.2.2/SPDX-license-expressions/#d3-simple-license-expressions
--- a/flit_core/flit_core/config.py
+++ b/flit_core/flit_core/config.py
@@ -850,7 +850,7 @@ def normalise_compound_license_expr(s: str) -> str:
                 parts.append(f' {part} ')
             elif part.lower() in {'and', 'or', 'with'}:
                 # provide a sensible error message for non-uppercase operators
-                reason = f"operators must be uppercase, not '{s}'"
+                reason = f"operators must be uppercase, not '{part}'"
                 raise ConfigError(invalid_msg.format(s=s, reason=reason))
             elif part == '(':
                 if parts and parts[-1] not in {' AND ', ' OR ', '('}:
--- a/flit_core/flit_core/config.py
+++ b/flit_core/flit_core/config.py
@@ -4,7 +4,6 @@ import errno
 import logging
 import os
 import os.path as osp
-from os.path import isabs
 from pathlib import Path
 import re
 
@@ -18,6 +17,7 @@ except ImportError:
     except ImportError:
         import tomli as tomllib
 
+from ._spdx_data import licenses
 from .common import normalise_core_metadata_name
 from .versionno import normalise_version
 
@@ -587,7 +587,8 @@ def read_pep621_metadata(proj, path) -> LoadedConfig:
     if 'license' in proj:
         _check_types(proj, 'license', (str, dict))
         if isinstance(proj['license'], str):
-            md_dict['license_expression'] = normalize_license_expr(proj['license'])
+            licence_expr = proj['license']
+            md_dict['license_expression'] = normalise_compound_license_expr(licence_expr)
         else:
             license_tbl = proj['license']
             unrec_keys = set(license_tbl.keys()) - {'text', 'file'}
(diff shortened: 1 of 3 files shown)


cdce8p (changes requested):
Thanks for working on this @AA-Turner! I like the approach you used here. Left a few suggestions.

flit_core/tests_core/test_config.py
```diff
@@ -222,8 +222,17 @@ def test_bad_pep621_readme(readme, err_match):
     ("mit",  "MIT"),
     ("apache-2.0", "Apache-2.0"),
     ("APACHE-2.0+", "Apache-2.0+"),
-    # TODO: compound expressions
-    #("mit and (apache-2.0 or bsd-2-clause)", "MIT AND (Apache-2.0 OR BSD-2-Clause)"),
+    ("mit AND (apache-2.0 OR bsd-2-clause)", "MIT AND (Apache-2.0 OR BSD-2-Clause)"),
+    ("(mit)", "(MIT)"),
```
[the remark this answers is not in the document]
AA-Turner: I thought about this but decided not to -- partly for simplicity, but also as I think Flit should do minimal processing of metadata. Currently all that happens is we normalise capitalisation (& spacing), and I could even see an argument to preserve the user's original whitespace.

A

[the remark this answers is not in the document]
cdce8p: Yeah this one is debatable. In the end it's probably not even worth considering as I hope the majority of users won't ever use that.

flit_core/tests_core/test_config.py
```diff
@@ -247,6 +257,10 @@ def test_license_expr_error():
     with pytest.raises(config.ConfigError, match="recognised"):
         config.read_pep621_metadata(proj, samples_dir / 'pep621' / 'pyproject.toml')
 
+    proj['license'] = "MIT WITH MIT-Exception"  # WITH unsupported
+    with pytest.raises(config.ConfigError, match="recognised"):
+        config.read_pep621_metadata(proj, samples_dir / 'pep621' / 'pyproject.toml')
```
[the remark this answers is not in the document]
AA-Turner: Many tests added

flit_core/flit_core/config.py
```diff
@@ -822,34 +823,79 @@ def isabs_ish(path):
...
 
 
-def normalize_license_expr(s: str):
-    """Validate & normalise an SPDX license expression
+def normalise_compound_license_expr(s: str):
+    """Validate and normalise a compund SPDX license expression.
 
-    For now this only handles simple expressions (referring to 1 license)
+    Per the specification, licence expression operators (AND, OR and WITH)
+    are matched case-sensitively. The WITH operator is not currently supported.
```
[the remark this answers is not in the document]
AA-Turner: I've added an error message & tests for lowercase operators, but I'll leave this unresolved for Thomas to comment.

My opinion on this is similar to the other comment, in that I think we should be doing minimal 'magic' here. From the PEP:

> #### [Add string value to `license` key](https://peps.python.org/pep-0639/#add-string-value-to-license-key)
>
> ...
>
> Build tools SHOULD validate and perform case normalization of the expression as described in the [Add License-Expression field](https://peps.python.org/pep-0639/#spec-field-license-expression) section, outputting an error or warning as specified.


> #### [Add `License-Expression` field](https://peps.python.org/pep-0639/#add-license-expression-field)
>
> The License-Expression optional [Core Metadata field](https://packaging.python.org/en/latest/glossary/#term-Core-Metadata-Field) is specified to contain a text string that is a valid SPDX [license expression](https://peps.python.org/pep-0639/#term-license-expression), as [defined above](https://peps.python.org/pep-0639/#spdx).
>
> Build and publishing tools SHOULD check that the License-Expression field contains a valid SPDX expression, including the validity of the particular license identifiers (as [defined above](https://peps.python.org/pep-0639/#spdx)). Tools MAY halt execution and raise an error when an invalid expression is found. If tools choose to validate the SPDX expression, they also SHOULD store a case-normalized version of the License-Expression field using the reference case for each SPDX license identifier and uppercase for the AND, OR and WITH keywords. Tools SHOULD report a warning and publishing tools MAY raise an error if one or more license identifiers have been marked as deprecated in the [SPDX License List](https://spdx.org/licenses/).

From this, Flit SHOULD check the validity of the expression (which this PR does), SHOULD use the correct-case version for the `License-Expression` field (which this PR does), and MAY raise an error when an invalid expression is found (which this PR chooses to do instead of coercing to uppercase).

Notably, none of the SPDX examples in the PEP use lowercase licence operators. It is unhelpfully ambiguous if a tool 'should' convert lowercase operators to uppercase, though (even for such a long PEP!).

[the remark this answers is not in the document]
cdce8p: I agree the PEP is ambiguous here. Providing a clear error message goes a long way to make the user experience better.

It's probably a judgement call whether or not Flit should allow lower case operators. I think it'd be fine if it does, then again it isn't that much work to capitalize them.

cdce8p (approved):
Overall this looks good to me! Thanks again @AA-Turner 👍🏻

flit_core/flit_core/config.py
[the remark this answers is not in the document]
takluyver: Thanks both. I also agree that it's somewhat ambiguous what the PEP intends, but given that  the SPDX spec says that the operators are case sensitive, I think sticking with that and emitting a clear error message is the way to go.

flit_core/flit_core/config.py
```diff
@@ -822,34 +824,85 @@ def isabs_ish(path):
...
+                if parts and parts[-1] not in {' AND ', ' OR ', '('}:
+                    raise ConfigError(f"{s!r} is not a recognised SPDX license ID")
+                stack += 1
+                parts.append(part)
+            elif part == ')':
+                if not parts or parts[-1] in {' AND ', ' OR ', ' WITH ', '('}:
+                    raise ConfigError(f"{s!r} is not a recognised SPDX license ID")
+                stack -= 1
+                if stack < 0:
+                    raise ConfigError(f"{s!r} is not a recognised SPDX license ID")
```
takluyver, line 864: 'License ID' was meant to refer to a single name (the SPDX spec calls these 'license identifiers'). For errors related to the structure of a compound expression, I'd use the term 'license expression', unless we can think of a clearer term for users.

Given that we've already got separate code paths for different error cases, I think it would be useful for the error message to try to describe why. Something like:

```
ConfigError: Invalid SPDX license expression (')' does not follow license ID): '(MIT AND)'
```

takluyver:
```
ConfigError: Invalid SPDX license expression: ')' does not follow license ID
    '(MIT AND)'

ConfigError: '(MIT AND)' is not a valid SPDX license expression: ')' does not follow license ID
```

Just trying out some different formats, the extra `()` parentheses in my original suggestion made it a bit hard to scan.

takluyver:
I will try to find some time to review this in more detail, but I think the basic structure looks good, and I appreciate the lengthy list of test cases :+1: .

flit_core/flit_core/config.py
AA-Turner: @takluyver I've added a reason to each error messages, and in the process fixed two bugs -- one where `'MIT MIT'` was allowed (no operator) and one where `' '` was allowed (whitespace only).

flit_core/flit_core/config.py
```diff
@@ -822,34 +824,96 @@ def isabs_ish(path):
...
+    Spec: https://spdx.github.io/spdx-spec/v2.2.2/SPDX-license-expressions/
+    """
+    invalid_msg = "'{s}' is not a valid SPDX license expression: {reason}"
+    if not s or s.isspace():
+        raise ConfigError(f"The SPDX license expression must not be empty")
+
+    stack = 0
+    parts = []
+    try:
+        for part in filter(None, re.split(r' +|([()])', s)):
```
AA-Turner, line 842: If we wanted to preserve whitespace we can do so here -- [`re.split()`](https://docs.python.org/3/library/re.html#re.split) returns the text of all groups used in the capturing pattern.

flit_core/flit_core/config.py
```diff
@@ -822,34 +824,96 @@ def isabs_ish(path):
...
+                raise ConfigError(invalid_msg.format(s=s, reason=reason))
+            elif part == '(':
+                if parts and parts[-1] not in {' AND ', ' OR ', '('}:
+                    reason = f"'(' must follow either AND, OR, or '('"
+                    raise ConfigError(invalid_msg.format(s=s, reason=reason))
+                stack += 1
+                parts.append(part)
+            elif part == ')':
+                if not parts or parts[-1] in {' AND ', ' OR ', ' WITH ', '('}:
+                    reason = "')' must follow a license ID"
```
cdce8p, line 863:
```py
'(MIT AND)' is not a valid SPDX license expression: ')' must follow a license ID
```
Maybe the reason could be improved here. It's not so much about the `)` position but rather that the second license id is missing.

flit_core/flit_core/config.py
```diff
@@ -822,34 +824,96 @@ def isabs_ish(path):
...
+    stack = 0
+    parts = []
+    try:
+        for part in filter(None, re.split(r' +|([()])', s)):
+            if part.upper() == 'WITH':
+                # provide a sensible error message for the WITH operator
+                raise ConfigError(f"The SPDX 'WITH' operator is not yet supported!")
+            elif part in {'AND', 'OR'}:
+                if not parts or parts[-1] in {' AND ', ' OR ', ' WITH ', '('}:
+                    reason = f"'{part}' must follow a license ID"
```
cdce8p, line 848:
```py
'AND MIT' is not a valid SPDX license expression: 'AND' must follow a license ID
```

flit_core/flit_core/config.py
```diff
@@ -822,34 +824,96 @@ def isabs_ish(path):
...
+                    reason = f"a license ID must follow either AND, OR, or '('"
+                    raise ConfigError(invalid_msg.format(s=s, reason=reason))
+                simple_expr = normalise_simple_license_expr(part)
+                parts.append(simple_expr)
+
+        if stack != 0:
+            reason = 'unbalanced brackets'
+            raise ConfigError(invalid_msg.format(s=s, reason=reason))
+        if parts[-1] in {' AND ', ' OR ', ' WITH '}:
+            reason = f"'{parts[-1].strip()}' is missing a second operand"
```
cdce8p, line 881:
```py
'MIT AND' is not a valid SPDX license expression: 'AND' is missing a second operand
```
What's missing here is the second license id not another operand.

flit_core/flit_core/config.py
cdce8p: I'd prefer to keep it. Even though the PEP only mentions `case-normalized`, removing unnecessary whitespaces improves readability IMO. `packaging` does that as well.

flit_core/flit_core/config.py
takluyver: Strictly the license ID is an operand, but I agree that the term probably isn't the most informative.

How about: "a license ID or expression should follow 'AND'"?

AA-Turner: I've rephrased to your suggestion.

flit_core/flit_core/config.py
AA-Turner: I've rephrased the error messages

flit_core/tests_core/test_config.py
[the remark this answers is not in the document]
takluyver: I could go for even more normalisation in the future - I think one meaning of normalisation is that if two expressions would be parsed the same way, their normalised form is also the same. But I agree people are unlikely to use superfluous parentheses, so :+1: for keeping it simple for now.

takluyver:
I've opened #735 to track the WITH operator, since we closed #723 and left that out.

takluyver:
Oh, and I meant to say: thanks for tackling this, I'm happy with how you've done it. :slightly_smiling_face:

PR #726, opened 2025-02-20
Formatting improvements

cdce8p:
Noticed these while working on other parts of the codebase. Use pre-commit checks to auto-fix it for the whole codebase
- Remove trailing whitespaces
https://github.com/pre-commit/pre-commit-hooks?tab=readme-ov-file#trailing-whitespace
- Fix end-of-file
https://github.com/pre-commit/pre-commit-hooks?tab=readme-ov-file#end-of-file-fixer

Added a basic `.pre-commit-config.yaml` file to make running them easier. I've stop short of enforcing it to not add an additional required development dependency.

PR #733, opened 2025-03-03
Set up Ruff to lint imports

AA-Turner:
From #732, I personally find it useful to have a tool check these things automatically for me.

This PR proposes a minimal Ruff configuration, setting up only unused import detection (F401) and import sorting. I've also taken the liberty to propose a GitHub workflow to check this.

However, I saw in @cdce8p's #726 that you "don't generally favour large-scale code cleanups or strict enforcement of style rules", so I won't be saddened if this PR gets rejected -- see it as just a suggestion of what might be useful!

A

---

N.B. There are several options to configure import sorting, e.g. [`force-sort-within-sections`](https://docs.astral.sh/ruff/settings/#lint_isort_force-sort-within-sections). I'm not sure if you have pre-existing preferences, so I just went with the isort defaults to begin with.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,33 @@
+name: Lint source code
+
+on:
+  push:
+  pull_request:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+env:
+  FORCE_COLOR: "1"
+
+jobs:
+  ruff:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        persist-credentials: false
+
+    - name: Install Ruff
+      uses: astral-sh/ruff-action@v3
+      with:
+        args: --version
+
+    - name: Lint with Ruff
+      run: ruff check --output-format=github
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,3 +4,12 @@ repos:
     hooks:
       - id: trailing-whitespace
       - id: end-of-file-fixer
+
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.9.9
+    hooks:
+      - id: ruff
+        name: "Lint with Ruff"
+        args:
+          - '--fix'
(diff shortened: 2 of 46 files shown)


cdce8p (commented):
Not sure about this one. I do prefer the imports to be sorted. However, I'm not sure `ruff` should be used for that. The `select` config does have the unfortunate consequence that more and more rules will be enabled as time goes one even if they might not be particularly relevant or even desirable.

Maybe a better approach would be to use `isort` as it's much more limited in scope? I'd also suggest to only add the `pre-commit` config for it.

If at some point, these should be enforced via CI, we could use https://pre-commit.ci for that.

--
One a related note, I'd personally also be open to adding a code formatter (preferably black) in a separate PR. Have to admit, when those were first added to one of the projects I work on, I was quite skeptical. However, over time it has grown on me and most could I write myself nowadays already follows these guidelines anyway.

AA-Turner:
> However, I'm not sure ruff should be used for that. The select config does have the unfortunate consequence that more and more rules will be enabled as time goes one even if they might not be particularly relevant or even desirable.

Could you elaborate? An explicit `select` as in this PR limits to just the isort rules and F401, nothing else will be added in updates of stuff unless explicitly opted in to. The benefit of using ruff over isort is that it is drastically faster, completing in less time than Python takes to start up!

A

cdce8p:
> Could you elaborate? An explicit `select` as in this PR limits to just the isort rules and F401, nothing else will be added in updates of stuff unless explicitly opted in to.

Of course. Was more thinking about what will happen in the future. I've seen it on multiple project now. First someone adds a limited set of carefully selected ruff rules. After some time, you'll get some random "drive-by" contributor who sees that and thinks they can help contribute to a project by enabling additional rules. It just tends to inflate a lot with time, unfortunately.

> The benefit of using ruff over isort is that it is drastically faster, completing in less time than Python takes to start up!

Yes, though isort should still be fast enough here. Especially as pre-commit isn't required. So enabling it is fully optional.

takluyver:
Thanks for the interest, but I am going to decline this for precisely the reason you mentioned - I personally prefer less tooling and less rigid rules about code style, and I don't like reformatting PRs changing lots of code.

If I do get persuaded on something like sorting imports,  I also have a rather untrendy preference for Python developer tools that are written in Python - like isort, unlike ruff - so we can readily understand and modify them without having to learn Rust.

PR #734, opened 2025-03-03
Use default `license-files` in `flit init`

AA-Turner:
The default `license-files` glob is `['COPYING*', 'LICEN[CS]E*']`, which is broader than what's currently set in `flit init`, importantly permitting file extensions. I think it makes sense to use the default glob here?

A


flit/init.py
```diff
@@ -224,7 +224,6 @@ def initialise(self):
                 f.write(tomli_w.dumps({'readme': readme}))
             if license != 'skip':
                 f.write(tomli_w.dumps({'license': license_names_to_spdx[license]}))
-                f.write(f"license-files = {json.dumps(['LICENSE'])}\n")
```
cdce8p, line 227: Adding a dedicated `license-files = ['LICENSE']` line to `pyproject.toml` does have the benefit that it's directly visible to the user. No need to lookup what the exact key is.

I'd suggest to keep it as is.

AA-Turner: Fair enough, my main complaint was that this didn't allow for an extension, so perhaps?:

cdce8p: > Fair enough, my main complaint was that this didn't allow for an extension

Does it need to though? I believe the value here is designed to match the initial config. If the license file name is changed, flit would also emit an error as the pattern didn't match any file, which would lead the user to update it.

AA-Turner:
Closing, this PR had marginal value, and justifiable concerns have been raised.

A

PR #738, opened 2025-03-25
Drop Python 3.6 and 3.7 support in flit-core

### LINKED ISSUE #730: Python version support
> We're probably coming to the point where even for flit_core I'll drop support for some older Python versions, but let's work that out separately. I generally find it easier to support a pretty broad range of Python versions than to deal with people getting stuck on an old version of a package I maintain because I dropped support for older Python.

_Originally posted by @takluyver in https://github.com/pypa/flit/pull/728#discussion_r1967261493_

Flit 3.11 supports Python 3.8+, and flit-core supports 3.6+.

Latest statistics [1] for flit-core downloads for show that 3.6 represents <0.05%, 3.7 represents <0.1%, and 3.8 <2%. 3.9 is the overwhelming majority at >50%.

For flit [2], Python 3.9 is again the overwhelming majority with >80%. The numbers are a little suspicious as 3.9 specifically skyrockets from late October/November onwards. I will try to look at a more granular level.

This probably suggests that dropping 3.6 and 3.7 for flit-core is a reasonable move, but other than that Flit should maintain support for non end-of-life Python releases.

A

[1]: https://pypistats.org/packages/flit-core
[2]: https://pypistats.org/packages/flit

AA-Turner:
Resolves #730.

Split into individual commits, feel free to push to this branch to drop any commits you don't want to take.

Supporting Python 3.8+ means that *flit* can use dataclasses, which might be nice, but I've stuck to small simplifications in this series.

A

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -51,35 +51,9 @@ jobs:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
         run: codecov
 
-  test-py37:
-    runs-on: "ubuntu-22.04"
-    strategy:
-      matrix:
-        python-version: [ "3.7", ]
-    steps:
-      - uses: actions/checkout@v4
-
-      - name: Setup Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v5
-        with:
-          python-version: ${{ matrix.python-version }}
-
-      - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip
-          pip install tox tox-gh-actions codecov
-
-      - name: Run tests
-        run: tox
-
-      - name: Codecov upload
-        env:
-          CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
-        run: codecov
-
   packages:
     runs-on: ubuntu-latest
-    needs: [test, test-py37]
+    needs: [test]
     permissions:
       id-token: write  # OIDC for uploading to PyPI
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -77,36 +77,9 @@ jobs:
           CODECOV_TOKEN: ${{ secrets.CODECOV_TOKEN }}
         run: codecov
 
-  test-py36:
-    runs-on: "ubuntu-20.04"
-    strategy:
-      matrix:
-        python-version: [ "3.6", ]
-    steps:
-      - uses: actions/checkout@v4
-
-      - name: Setup Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v5
-        with:
-          python-version: ${{ matrix.python-version }}
-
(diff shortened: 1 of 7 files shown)


cdce8p:
Thanks for the PR! Have we reached a consensus on #730 yet? I'd be sympathetic to dropping 3.6 and 3.7 but maybe we should formalize that a bit first.

As for the PR itself, I would usually prefer to split it up a bit differently. E.g. start with dropping the Python versions, then refactoring and typing changes in followup PRs but that might be just my opinion.

takluyver:
I might split it up too, but doing one PR does have the nice feature of showing what simplifications we immediately gain access to by dropping old Python versions. At least, in principle - I think most of the changes here are actually leftover from when flit_core supported Python 2 for a bit, which I never went back and cleaned up. :slightly_smiling_face: 

Is there any particular benefit to sprinkling `from __future__ import annotations` on every module? It looks like there's one place where it's letting us write an annotation in a neater way, but in most files there's no change to the actual annotations?

AA-Turner:
> At least, in principle - I think most of the changes here are actually leftover from when flit_core supported Python 2 for a bit, which I never went back and cleaned up.

I realised this whilst preparing the PR, hence why the first five commits are before dropping support for Python 3.6. Apologies for not making this clearer in the description!

A

takluyver:
No worries. I'd lean towards adding the `__future__` import as and when it's useful, rather than in anticipation, but I also don't feel strongly about that.

On type annotations more broadly, I like them as progressive hints where they improve the clarity of code - indicating that a path parameter is a `Path` object rather than a `str` is definitely valuable - but less so when every function signature has to be spread over multiple lines to accommodate `TypeVar`s and `Literal`s. I am clearly out of step with the consensus that type hints are great and we should be using them wherever possible, but please don't do aim to add exhaustive type hints on the basis that more is always better. :wink: 

I also appreciate the argument that type hints on APIs are useful to code using those, but the only part of Flit anyone else should be importing & calling is `flit_core.buildapi`, so this argument doesn't apply to most of our code.

AA-Turner:
> the consensus that type hints are great and we should be using them wherever possible

Likewise I'm sceptical. I am also wary of the cost of annotations -- both in performance and maintainability. However, over time I have found them useful in Sphinx, as they enable local reasoning about a particular function, attribute, etc, without having to resort to looking up the definition.

To return to this PR, would you prefer I split it into three parts (general clean-up, remove 3.6, remove 3.7)? To reduce the amount of rebasing etc work, I'd selfishly prefer to keep it as one PR, but I'm happy to split if that makes review/etc easier.

A

takluyver:
No, leave it as one, I think it's manageable to review as is. But if you're happy to remove the `__future__` imports for now where they're not actually making any difference, go for it.

takluyver:
Thanks, this looks good. :+1:

PR #739, opened 2025-03-25
Update the bootstrap guide for ``tomllib``

AA-Turner:


doc/bootstrap.rst
```diff
@@ -26,9 +26,12 @@ included in the sdist (or by unzipping it to the correct directory)::
     # Install somewhere else:
     python bootstrap_install.py --installdir /path/to/site-packages dist/flit_core-*.whl
 
-As of version 3.6, flit_core bundles the ``tomli`` TOML parser, to avoid a
-dependency cycle. If you need to unbundle it, you will need to special-case
-installing flit_core and/or tomli to get around that cycle.
+In Python 3.11 and newer, ``flit_core`` uses the standard library :mod:`tomllib`
+module, meaning that ``flit_core`` has no third-party dependencies.
```
cdce8p, line 30: It's vendored for all Python versions, just not used in `>=3.11`.

cdce8p (commented):
This seems overly complicated to me. I'd assume that redistributors will need to check how `tomli` themselves anyway, so I don't see a particular need or improvement with this change.

doc/bootstrap.rst
```diff
@@ -26,9 +26,18 @@ included in the sdist (or by unzipping it to the correct directory)::
...
+.. note::
+
+   This note only applies if you need to unbundle or unvendor dependencies.
+
+   ``flit_core`` unconditionally bundles the ``tomli`` TOML parser,
+   to avoid a dependency cycle for older versions of Python that don't include
+   the standard library :mod:`tomllib` module.
+
+   The ``tomli`` library is only used on Python 3.10 and older, meaning that
+   if you use Python 3.11 or newer you can safely unbundle ``tomli``.
```
takluyver, line 38:
```suggestion
   if you use Python 3.11 or newer you can simply remove the bundled ``tomli``.
```

takluyver:
Thanks, that looks fine to me. :+1:

PR #740, opened 2025-03-27
Use f-strings

AA-Turner:
Beyond #738, this PR seeks to consistently adopt f-strings in both flit and flit-core. I realise this PR has a reasonably large diff, but I would argue that f-strings are both more readable and have better performance than either printf-style formatting or `str.format()` calls.

A


cdce8p (approved):
Thanks @AA-Turner 👍🏻 I agree that f-strings are almost always the better choice. The PR LGTM!

AA-Turner:
@cdce8p are you now a flit maintainer? Congratulations!

A

PR #742, opened 2025-04-06
feat: allow stubs package

### LINKED ISSUE #332: Support stub-only packages?
When trying to use flit to build a [PEP 561 stub-only package](https://www.python.org/dev/peps/pep-0561/#stub-only-packages), I found two main blockers:

* flit requires that `module` is a valid identifier, which I think is not expected for these packages (to be named `*-stubs`)
* flit looks for a `__init__.py`, which is not there

I understand that perhaps you want flit to support only normal Python packages, but I still wanted to open this feature request just in case :)

trim21:
close https://github.com/pypa/flit/issues/332

- do not try to get docstring and version if it's not `dynamic`, which is typical the case of stubs package
- do not raise exception if it's `-stubs` package without `__init__.py`
- do not normalize the `-stubs` parts in package name


flit_core/flit_core/config.py
```diff
@@ -510,6 +510,12 @@ def _check_list_of_str(d, field_name):
             "{} field should be a list of strings".format(field_name)
         )
 
+def normalize_pkg_name(name:str)->str:
+    if name.endswith('-stubs'):
+        return name.removesuffix('-stubs').replace('-','_') + '-stubs'
```
AA-Turner, line 515: The special casing should be explained here. For example, typeshed uses a `types-` prefix, which this doesn't cover. Is that important?

trim21: it's from pep-0561, a type stub package should add `-stubs` to main package.

the `types-` prefix is just the pypi name, which is not the director contains pyi files.

For example, if you unzip the `types_six-1.17.0.20250403-py3-none-any.whl`, you will get this:

```text
Archive:  types_six-1.17.0.20250403-py3-none-any.whl
  inflating: six-stubs/METADATA.toml
  inflating: six-stubs/__init__.pyi
  inflating: six-stubs/py.typed
  inflating: six-stubs/moves/BaseHTTPServer.pyi
  inflating: six-stubs/moves/CGIHTTPServer.pyi
  inflating: six-stubs/moves/SimpleHTTPServer.pyi
  inflating: six-stubs/moves/__init__.pyi
  inflating: six-stubs/moves/_dummy_thread.pyi
...
```

trim21: It's still possible that a package named as something like `pkg-stubs2` and contins `pkg-stubs/__init__.pyi` for package "pkg", but I don't think it's worth handling ...

flit_core/flit_core/sdist.py
```diff
@@ -144,8 +144,9 @@ def apply_includes_excludes(self, files):
         crucial_files = set(
             self.extra_files + [str(self.module.file.relative_to(self.cfgdir))]
         )
+
         missing_crucial = crucial_files - files
-        if missing_crucial:
+        if missing_crucial and not self.module.name.endswith('-stubs'):
```
cdce8p, line 149: A check like this feels wrong. Ideally, `__init__.py` wouldn't even be added to `crucial_files` for stub packages. Furthermore, it might make sense to do the `endswith("-stubs")` check once and store it if it's a stubs package in the `flit.common.Module` so we can access it here, e.g. with `self.module.is_stubs_pkg`.

flit_core/flit_core/buildapi.py
```diff
@@ -24,11 +24,15 @@ def get_requires_for_build_wheel(config_settings=None):
     # by parsing the module (_via_ast), we don't need any extra
     # dependencies. If not, we'll need to try importing it, so report any
     # runtime dependencies as build dependencies.
+
```
cdce8p, line 27:
```suggestion
```
The empty line here seems unnecessary.

cdce8p (changes requested):
Thanks for working on this @trim21! It will be a great addition to have. Left a few comments below.

flit_core/flit_core/config.py
cdce8p: Just tested this branch and `types-pkg` would also works as a package name (if `tool.flit.module` is used). The only important thing for PEP 561 as I recall is that the package dir name is `{original_name}-stubs`. E.g. `foopkg-stubs` for `foopkg`.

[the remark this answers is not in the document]
trim21: good catch, forget we still support python 3.8

flit_core/flit_core/sdist.py
trim21: > A check like this feels wrong. Ideally, __init__.py wouldn't even be added to crucial_files for stub packages.

that's why I skip the check here, if it's a stub package.

> Furthermore, it might make sense to do the endswith("-stubs") check once and store it if it's a stubs package in the flit.common.Module so we can access it here, e.g. with self.module.is_stubs_pkg.

make sence

flit_core/flit_core/common.py
```diff
@@ -18,8 +18,11 @@ class Module:
     in_namespace_package = False
     namespace_package_name = None
 
-    def __init__(self, name, directory=Path()):
+    is_stub_pkg: bool
+
+    def __init__(self, name: str, directory=Path()):
         self.name = name
+        self.is_stub_pkg = name.endswith('-stub')
```
cdce8p, line 25:
```suggestion
    def __init__(self, name: str, directory=Path()):
        self.name = name
        self.is_stub_pkg = name.endswith('-stub')
```

No need to add the explicit annotation here. Type checkers will be able to infer it.

flit_core/flit_core/sdist.py
```diff
@@ -73,7 +73,7 @@ class SdistBuilder:
     """
     def __init__(self, module, metadata, cfgdir, reqs_by_extra, entrypoints,
                  extra_files, data_directory, include_patterns=(), exclude_patterns=()):
-        self.module = module
+        self.module: common.Module = module
```
cdce8p, line 76:
```suggestion
        self.module = module
```
The annotation should be added to the `__init__` parameter at some point. I'd defer that to some time in the future as it would require reformatting the whole lines.

flit_core/flit_core/sdist.py
```diff
@@ -144,8 +144,9 @@ def apply_includes_excludes(self, files):
         crucial_files = set(
             self.extra_files + [str(self.module.file.relative_to(self.cfgdir))]
         )
+
```
cdce8p, line 147:
```suggestion
```
Empty line

cdce8p (commented):
Some last comments

flit_core/flit_core/common.py
trim21: at least pyright can't? I didn't get type info in vscode

flit_core/flit_core/sdist.py
trim21: I don't think they have same semantics when you add it to `__init__` parameter or `self.attr` assign. But the type annotation is not the point of this PR, if you disagree with it I'll just remove it.

flit_core/flit_core/common.py
cdce8p: > at least pyright can't? I didn't get type info in vscode

It works once `name` is annotated as `str`.

Screenshot 2025-05-07 at 12 41 40

flit_core/flit_core/sdist.py
cdce8p: > I don't think they have same semantics when you add it to `__init__` parameter or `self.attr` assign.

Strictly speaking you're right. It's just that IMO attribute annotations don't add much value if they are identical as the parameter ones. I know that's one area where type checkers (in particular mypy and pyright) disagree though.

_In my own experience it works good enough with just parameter annotations._

cdce8p (changes requested):

flit_core/flit_core/sdist.py
```diff
@@ -145,7 +145,7 @@ def apply_includes_excludes(self, files):
             self.extra_files + [str(self.module.file.relative_to(self.cfgdir))]
         )
         missing_crucial = crucial_files - files
-        if missing_crucial:
+        if missing_crucial and not self.module.is_stub_pkg:
```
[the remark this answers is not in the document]
trim21: is this what you have in mind?

flit_core/flit_core/common.py
```diff
@@ -18,8 +18,9 @@ class Module:
     in_namespace_package = False
     namespace_package_name = None
 
-    def __init__(self, name, directory=Path()):
+    def __init__(self, name: str, directory=Path()):
         self.name = name
+        self.is_stub_pkg = name.endswith('-stub')
```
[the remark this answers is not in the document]
trim21: Opps, good catch. it's `stubs`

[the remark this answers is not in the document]
cdce8p: > Shouldn't this be `-stubs`? I do wonder why the test didn't fail. The package doesn't contain an `__init__.py` file.

The check only fails during the `sdist` build. The test case only builds the wheel. It would probably make sense to add another test case in `tests/test_sdist.py` then.

flit_core/flit_core/config.py
trim21: Actually any package name works, if `tool.flit.module` is given. this is how I build `foopkg-stubs`  with flit-core currently.

cdce8p (approved):
Thanks a lot @trim21! This looks good to me. Will wait a few more days before merging to give others the chance to comment.

cdce8p (approved):
@trim21 Sorry for the long delay. Will go ahead and merge this now.

PR #748, opened 2025-07-09
Describe more clearly what gets included in distribution files

pfmoore:


doc/pyproject_toml.rst
```diff
@@ -399,19 +399,23 @@ They are in the same format as the newer ``[project.entry-points]`` tables
...
 
-If you want more control, you can give lists of paths or glob patterns as
-``include`` and ``exclude`` in this section. For example:
+When building an sdist, in addition to the package contents Flit includes
+everything it needs to build and install your module: your ``pyproject.toml``
+file, the readme & license files given in the metadata, and the :ref:`external
+data folder <pyproject_toml_external_data>` if you specified that.
+
+If you want more control over the content of the sdist, you can give lists of
+paths or glob patterns as ``include`` and ``exclude`` in the ``sdist`` section
```
takluyver, line 417:
```suggestion
paths or glob patterns as ``include`` and ``exclude`` in the ``tool.flit.sdist`` section
```

doc/pyproject_toml.rst
```diff
@@ -399,19 +399,23 @@ They are in the same format as the newer ``[project.entry-points]`` tables
...
+metadata.
 
-If you want more control, you can give lists of paths or glob patterns as
-``include`` and ``exclude`` in this section. For example:
+When building an sdist, in addition to the package contents Flit includes
+everything it needs to build and install your module: your ``pyproject.toml``
+file, the readme & license files given in the metadata, and the :ref:`external
+data folder <pyproject_toml_external_data>` if you specified that.
+
+If you want more control over the content of the sdist, you can give lists of
```
takluyver, line 416: While the include & exclude lists are only used when building the sdist, the exclude list can affect wheels as well, because the common pattern (including in `build`) is to build the sdist first, unpack it, and build a wheel from that.

I'm not entirely sure of the best way to include this information, so feel free to leave it for me to address separately, but I thought it was worth mentioning while we're looking at this.

takluyver (commented):
Thanks, this looks reasonable.

doc/pyproject_toml.rst
pfmoore: I didn't explicitly point out that the *include* list can't affect the wheel (as it only adds non-package content), because I felt things could get too complicated if I did so. But again, let me know if you disagree.

PR #750, opened 2025-07-09
Generate legal `RECORD` files when file names contain commas

### LINKED ISSUE #744: [bug] flit_core will generate illegal RECORD
According to
https://github.com/pypa/auditwheel/issues/587#issuecomment-2954217431

https://github.com/pypa/flit//blob/main/flit_core/flit_core/wheel.py#L198
is incorrect.

A possible solution is:

```diff
diff --git a/flit_core/flit_core/wheel.py b/flit_core/flit_core/wheel.py
index a3b752c..cdbb6ae 100644
--- a/flit_core/flit_core/wheel.py
+++ b/flit_core/flit_core/wheel.py
@@ -195,6 +195,8 @@ class WheelBuilder:
         # Write a record of the files in the wheel
         with self._write_to_zip(f'{self.dist_info}/RECORD') as f:
             for path, hash, size in self.records:
+                if path.find(',') >= 0:
+                    path = f'"{path}"'
                 f.write(f'{path},sha256={hash},{size}\n')
             # RECORD itself is recorded with no hash or size
             f.write(f'{self.dist_info}/RECORD,,\n')
```

SpecLad:
Fixes #744.

### DIFF
--- a/flit_core/flit_core/wheel.py
+++ b/flit_core/flit_core/wheel.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 import argparse
 from base64 import urlsafe_b64encode
 import contextlib
+import csv
 from datetime import datetime, timezone
 import hashlib
 from io import StringIO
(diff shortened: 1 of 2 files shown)


flit_core/flit_core/wheel.py
```diff
@@ -138,8 +139,8 @@ def _add_file(self, full_path, rel_path):
         self.records.append((rel_path, hash_digest, size))
 
     @contextlib.contextmanager
-    def _write_to_zip(self, rel_path, mode=0o644):
-        sio = StringIO()
+    def _write_to_zip(self, rel_path, mode=0o644, newline='\n'):
+        sio = StringIO(newline=newline)
```
takluyver, line 143: I think passing `newline` around is not needed here; I see the csv module docs say to use `newline=''`, but `'\n'` (the default for StringIO) is equivalent:

> When writing output to the stream... If newline is '' or '\n', no translation takes place.

takluyver:
Thanks, I think we can simplify it slightly, but otherwise this looks good. :+1:

flit_core/flit_core/wheel.py
SpecLad: Oh yeah, that's a good point. Reverted this part.

takluyver:
Thanks!

Issue #749, opened 2025-07-09
License files that are inside the source folder get included twice


SpecLad:
Create a project like this:

`pyproject.toml`:

```toml
[build-system]
requires = ["flit_core >=3.11,<4"]
build-backend = "flit_core.buildapi"

[project]
name = "foobar"
version = "0"
description = "bazquux"
license-files = ["foobar/test.txt"]
```

`foobar/__init__.py`: empty
`foobar/test.txt`: empty

Build it:

```console
> pyproject-build
```

Examine the results:

```console
> tar -tf .\dist\foobar-0.tar.gz
foobar-0/foobar/test.txt <---
foobar-0/foobar/__init__.py
foobar-0/foobar/test.txt <---
foobar-0/pyproject.toml
foobar-0/PKG-INFO
> tar -tf .\dist\foobar-0-py2.py3-none-any.whl
foobar/__init__.py
foobar/test.txt <---
foobar-0.dist-info/licenses/foobar/test.txt <---
foobar-0.dist-info/WHEEL
foobar-0.dist-info/METADATA
foobar-0.dist-info/RECORD
```

As you can see, in both the sdist and the wheel, the license file is included twice. The sdist problem I was only able to reproduce on Windows. The wheel problem occurs on both Windows and Linux.

takluyver:
The wheel problem is, in my view, not a problem. As was just made clearer in the docs (#748), all files within the package directory - `foobar/` in your example - are included by default. And they're included in the same relative locations as they are in the source folder (this isn't explicitly stated, but as we don't offer any way to rearrange them, it's the only logical outcome). That's the `foobar/test.txt` copy. Then files indicated by `license-files` are included in `...dist-info/licenses` [like the specification says](https://packaging.python.org/en/latest/specifications/binary-distribution-format/#the-dist-info-licenses-directory).

If you don't want the first copy, I think listing it in `excludes` and building with `pyproject-build` should work. If it doesn't, that's a bug. If you don't want the second copy, don't put it in `license-files`.

The sdist issue does seem like a bug, probably something to do with path normalisation on Windows.

SpecLad:
> On second thoughts, I'm also going to claim the error you see on Linux makes sense

That's fair enough, but then we're back to the original problem - there's no way to put the license only in the `*.dist-info/licenses` directory.

Do you think flit should provide some way to accomplish that, or is it out of scope?

PR #752, opened 2025-07-11
Fix license file duplication in sdists

SpecLad:
This bug can happen on Windows when the `license-files` setting contains a glob that matches one of the files in the source directory. In this case, the same path can be returned by `select_files` both in the POSIX format and the Windows format, which makes the deduplication logic in `apply_includes_excludes` fail.

Fix it by making sure that all paths passed to `apply_includes_excludes` are in the native OS format.

Fixes half of #749.


takluyver:
Thanks, and in particular thanks for writing a test as well. :+1: Did you check that the test fails on Windows without the fix? If not, I'll make a draft PR reverting the fix to check that in CI.

PR #756, opened 2025-07-31
Use raw string

DimitriPapadopoulos:
As in upstream distlib:
https://github.com/pypa/distlib/commit/90e6da4

### DIFF
--- a/flit_core/flit_core/common.py
+++ b/flit_core/flit_core/common.py
@@ -259,14 +259,14 @@ def check_version(version):
     return version
 
 
-script_template = """\
+script_template = r"""\
 #!{interpreter}
 # -*- coding: utf-8 -*-
 import re
 import sys
 from {module} import {import_name}
 if __name__ == '__main__':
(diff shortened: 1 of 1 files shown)


flit_core/flit_core/common.py
```diff
@@ -259,14 +259,14 @@ def check_version(version):
     return version
 
 
-script_template = """\
+script_template = r"""\
```
takluyver, line 262: Thanks, but the regular string is necessary for the `"""\<newline>` to work correctly. I think this can stay as is.

DimitriPapadopoulos: Ah indeed, upstream doesn't use `"""\`. I thought you wanted to keep more or less in sync with upstream because of #359:
> Update generated scripts to match pip installed scripts

But it's about matching the generated scripts, not necessarily the code that generates them.

takluyver: Yup, that sounds about right. I like having the script contents starting at the start of a line, and this seems like a neat way to do it.

PR #758, opened 2025-08-03
Default `license-files` glob: consistency between backends

### LINKED ISSUE #757: Make `license-files` default value consistent between backends?
### flit
https://github.com/pypa/flit/blob/7d9a070a44f7516eab6095505d80d93f13320388/flit_core/flit_core/config.py#L77

### setuptools

**[setuptools/setuptools/dist.py](https://github.com/pypa/setuptools/blob/3544de73b3662a27fac14d8eb9f5c841668d66de/setuptools/dist.py#L594-L597)**
Lines 594 to 597 in [3544de7](https://github.com/pypa/setuptools/commit/3544de73b3662a27fac14d8eb9f5c841668d66de):
```python
            # Default patterns match the ones wheel uses
            # See https://wheel.readthedocs.io/en/stable/user_guide.html
            # -> 'Including license files in the generated wheel file'
            patterns = ['LICEN[CS]E*', 'COPYING*', 'NOTICE*', 'AUTHORS*']
```

### hatch

**[hatch/backend/src/hatchling/metadata/core.py](https://github.com/pypa/hatch/blame/28f233c535508247ffa9a40dab82c1d1cb2b700b/backend/src/hatchling/metadata/core.py#L746)**
Line 749 in [28f233c](https://github.com/pypa/hatch/commit/28f233c535508247ffa9a40dab82c1d1cb2b700b)
```python
                globs = ['LICEN[CS]E*', 'COPYING*', 'NOTICE*', 'AUTHORS*']
```

DimitriPapadopoulos:
Fixes #757.

**Edit:** [Consistency with wheels](https://wheel.readthedocs.io/en/stable/user_guide.html#including-license-files-in-the-generated-wheel-file) is probably a better argument than consistency with other backends.

> Several open source licenses require the license text to be included in every distributable artifact of the project. By default, `wheel` conveniently includes files matching the following [glob](https://docs.python.org/library/glob.html) patterns in the `.dist-info` directory:
> * `AUTHORS*`
> * `COPYING*`
> * `LICEN[CS]E*`
> * `NOTICE*`

### DIFF
--- a/doc/pyproject_toml.rst
+++ b/doc/pyproject_toml.rst
@@ -101,7 +101,7 @@ license
   ``text`` key (the license text).
 license-files
   A list of glob patterns for license files to include.
-  Defaults to ``['COPYING*', 'LICEN[CS]E*']``.
+  Defaults to ``['COPYING*', 'LICEN[CS]E*', 'NOTICE*', 'AUTHORS*']``.
 authors
   A list of tables with ``name`` and ``email`` keys (both optional) describing
   the authors of the project.
--- a/flit_core/flit_core/config.py
+++ b/flit_core/flit_core/config.py
@@ -74,7 +74,7 @@ pep621_allowed_fields = {
     'dynamic',
 }
 
-default_license_files_globs = ['COPYING*', 'LICEN[CS]E*']
+default_license_files_globs = ['COPYING*', 'LICEN[CS]E*', 'NOTICE*', 'AUTHORS*']
 license_files_allowed_chars = re.compile(r'^[\w\-\.\/\*\?\[\]]+$')
 
 


takluyver:
I don't think backends necessarily need to align on things like this, but I also have no particular objection to the additions. Don't forget the docs need updating too.

DimitriPapadopoulos:
They don't **have** to align of course, but it would make the whole Python ecosystem more intelligible for me :smile:

takluyver:
Thanks - I'll give this a couple of days for other maintainers/contributors to take a look, and then merge if there are no objections. Feel free to remind me if I forget (which is fairly likely). :slightly_smiling_face:

cdce8p (approved):
Thought about this as well when I was working on the PEP 639 implementation. Decided not to include it there just to get the PR over the finish line earlier.

Overall it makes sense for the ecosystem as a whole to use similar globs though this isn't a strict requirement.

PR #759, opened 2025-08-03
Update license templates

DimitriPapadopoulos:
Get the Apache and GPL licenses directly from the sources:
* https://www.apache.org/licenses/LICENSE-2.0.txt
* https://www.gnu.org/licenses/gpl-3.0.txt

Remove the title from the MIT license, for consistency with:
* https://opensource.org/license/MIT
* https://fedoraproject.org/wiki/Licensing:MIT?rd=Licensing/MIT
* http://www.jclark.com/xml/copying.txt


DimitriPapadopoulos:
Ah, I hadn't noticed “_Text in italicized blue is omittable_”. Yet, I would prefer sticking to the text of the licence as provided by the FSF itself:
https://www.gnu.org/licenses/gpl-3.0.en.html

Most software I know of use the whole licence—perhaps without really having thought about it though.

With that said, I can remove the omittable end if you insist on it.

takluyver:
I would still prefer to cut off the 'how to apply' appendices, thanks. I think you're probably right that many projects copy the whole text without thinking about this, but now that we have thought about it, we don't have to follow that.

PR #777, opened 2025-09-23
Adds handling of tuples for version

skrawcz:
This change checks if the version is a tuple and then maps it to a string delimited by `.`. It then follows the natural path it does.

To use flit we need this for our [Apache project](https://github.com/apache/hamilton/blob/main/hamilton/version.py#L18).

### DIFF
--- a/flit_core/flit_core/common.py
+++ b/flit_core/flit_core/common.py
@@ -251,6 +251,8 @@ def check_version(version):
     if not version:
         raise NoVersionError('Cannot package module without a version string. '
                              'Please define a `__version__ = "x.y.z"` in your module.')
+    if isinstance(version, tuple):
+        version = '.'.join(map(str, version))
     if not isinstance(version, str):
         raise InvalidVersion(f'__version__ must be a string, not {type(version)}.')
 
--- a/flit_core/tests_core/test_common.py
+++ b/flit_core/tests_core/test_common.py
@@ -101,13 +101,15 @@ class ModuleTests(TestCase):
             check_version('3!')
 
         with pytest.raises(InvalidVersion):
-            check_version((1, 2))
+            check_version((1, 'a', 2))
 
         with pytest.raises(NoVersionError):
             check_version(None)
 
         assert check_version('4.1.0beta1') == '4.1.0b1'
         assert check_version('v1.2') == '1.2'
+        assert check_version((1, 2, 3)) == '1.2.3'
+        assert check_version((4,1,0, 'beta1')) == '4.1.0b1'
 
 def test_normalize_file_permissions():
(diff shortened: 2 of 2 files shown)


takluyver:
Thank-you, but sorry, this is not an option I want to allow.

The convention is that `mod.__version__` is a string. This predates Flit by some way, and one of my goals with Flit is to reinforce conventions like this, encouraging Python packages to be more uniform & predictable. Allowing `__version__` to be a tuple goes against that. I understand the advantages of representing version numbers as tuples, but strings are the norm in Python packaging, right down to the metadata formats and the [standard library](https://docs.python.org/3/library/importlib.metadata.html#importlib.metadata.version).

You could:

1. Convert `__version__` to a string and have another attribute e.g. `version_info` for a tuple, [like IPython](https://github.com/ipython/ipython/blob/036d1fd3d3c4afe239fe47e4a6030dd43ca3b946/IPython/core/release.py#L28-L33). The `version_info` name is based on `sys.version_info` for Python itself. This is the best option from my perspective. :slightly_smiling_face: 
2. Keep a copy of the version number in `pyproject.toml`, as I see your PR currently does.
3. Carry on using setuptools, if neither of those options are acceptable.

Again, sorry to say no, but I hope you understand my rationale.

skrawcz:
I disagree. Anyone who needs to do anything with semantic versions needs to parse it. So a tuple is just the already parsed version. It should be an allowed option. As you can see with this PR, your design allows for it to cleanly exist. Note: this only impacts existing projects that did not grow up with flit. If you feel strongly you can print a warning to guide folks to not do that...

For us I cannot change __version__ to be a string because it would not be backwards compatible. This is a barrier to adopting flit. Alternatively, does flit have an option to point to another parameter name? Since that would work as I could create another field which is the string version.

takluyver:
Python code handling version numbers will generally parse them using [`packaging.version.Version()`](https://packaging.pypa.io/en/stable/version.html), which is smarter about stuff like pre-releases than bare tuples. But that's not really important here. The convention is that version numbers are strings - this approach has its drawbacks, but we're used to working with them, and consistency is a big benefit.

> does flit have an option to point to another parameter name?

It doesn't, the only options are to specify the version in `pyproject.toml` or to have a string in `mod.__version__`. I'd decline a PR adding such an option on the same grounds - I value simplicity and consistency over flexibility in this.

> This is a barrier to adopting flit

In that case, possibly Flit doesn't work for you, sorry. It was never meant to serve every use case.

takluyver:
There are tools that can help with keeping the version number in sync, if you go that way. I've used [bump2version](https://github.com/c4urself/bump2version/) on some projects, though I see it's now recommending [bump-my-version](https://github.com/callowayproject/bump-my-version) as a replacement.

Issue #673, opened 2024-01-28
Flit 4 planning


takluyver:
What major changes do we want for Flit 4.0?

- [ ] Remove support for old-style metadata, where a project is described in the `[tool.flit]` subtables instead of the now-standard `[project]` table. (but `[tool.flit.module]`, which is used alongside `[project]`, will continue to work).
  - I encourage all projects to specify something like `requires = ["flit_core >=3.2,<4"]` so we can make changes like this and old packages can still easily be built with 3.x. I know some projects have ignored that and don't specify an upper bound, but hopefully not too many.
- [ ] Fail on unrecognised keys in `[project]` - #505, #506
- [ ] Get rid of `flit.buildapi` in favour of `flit_core.buildapi`? #519
- [ ] Switch the default behaviour of `flit build` and `flit publish` to not use VCS info, to get rid of the surprising difference between `flit build` and `python -m build` (see discussion on #522 & #625)
  - [ ] Make a tool to generate [include & exclude lists for sdists](https://flit.pypa.io/en/stable/pyproject_toml.html#sdist-section) from the VCS? Or a separate way to explicitly make a 'release sdist' by doing `git archive` or similar? :thinking: 

Anything else?

There's still likely to be a 3.10 release first with assorted small fixes and improvements, and possibly more 3.x releases before 4.0.

takluyver:
Should we get rid of the option to generate a setup.py file in sdists? We turned that off by default in 3.5 (November 2021).

liZe:
I like all your proposals. I’m still not sure about the best way to keep VCS support for build/publish, but having `flit build` and `python -m build` building the same package by default looks like a good idea.

Thanks for your work on Flit!

z3dev:
Complete keyring support.

takluyver:
I've opened #771 for the first two points of this - dropping `[tool.flit.metadata]` support and erroring on unknown keys in `[project]`.

takluyver:
And #772 would make `--no-use-vcs` the default, bringing `flit build` in line with `python -m build` using flit_core as a backend.

takluyver:
We're getting closer to Flit 4. I will wait for at least a new release of `packaging` first, because currently twine rejects packages with the latest metadata fields (#779), and I don't really want to revert that change.

#782 is a rough draft of the script to generate the include/exclude rules from git.

takluyver:
[...]

PR #782, opened 2025-10-25
Add script to generate sdist included/exclude patterns from git

takluyver:
This is meant to ease making `--no-use-vcs` the default for `flit build` and `flit publish`, by generating sdist include & exclude patterns to produce sdists with the same contents as before. See #673.

To use it:

- Generate any files that should be *excluded* from the sdist, e.g. virtualenvs, built docs, test coverage results, so the script can see them.
  - But only files/folders within included folders actually need to be excluded - extra things in the project root are left out by default
- `cd` to the project root (where `pyproject.toml` is)
- Run `python -m flit.sdist_rules`
- Copy the output into `pyproject.toml`, make any changes you want

At present, this only looks at the files in the folder, aiming to include anything tracked in git, and exclude anything that isn't. This is why it's necessary to generate the unwanted files first. I'm still thinking about whether it can be smarter using `.gitignore`.

### DIFF
--- a/doc/pyproject_toml.rst
+++ b/doc/pyproject_toml.rst
@@ -294,6 +294,14 @@ These paths:
 .. versionchanged:: 3.8
    Include and exclude patterns can now use recursive glob patterns (``**``).
 
+.. note::
+
+   Flit includes a tool to generate these rules for you from a git repository.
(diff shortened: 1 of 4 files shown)


takluyver:
This probably isn't perfect, but I think it's good enough.

PR #784, opened 2025-11-22
Add GitHub Actions workflow for Python package

Kelleretoro:
This workflow installs Python dependencies, runs tests, and lints code with multiple Python versions.


takluyver:
We already have a CI workflow set up. You appear to be submitting low-value pull requests like this to a wide range of projects, which is a form of spamming - it takes time and energy for maintainers to look at these and decide that they're not useful. And apparently Github now automatically burns resources generating AI reviews as well, whether I ask it to or not.

If you want to make useful contributions to open source projects, that's great, but you'll need to focus on some projects that you're interested in, and take the time to work out what might be useful.

PR #785, opened 2025-11-22
Remove flit.buildapi module

### LINKED ISSUE #519: Emit warnings when flit.buildapi is used
Could flit be made to output warnings when `flit.buildapi` backend is used? I think there's no reason to use it rather than `flit_core.buildapi`, is there?

takluyver:
Packages should use `flit_core.buildapi` as the PEP 517 backend, instead of `flit.buildapi`.

I know there are packages out there still using `flit.buildapi`, which existed first. All the ones I've seen are also still using older-style metadata (`[tool.flit.metadata]` tables instead of `[project]`), so they won't build with Flit 4.0 anyway. If anyone finds an example with `[project]` metadata but `flit.buildapi` as the backend before Flit 4.0, I'll leave this module around to avoid needless breakage.

Closes #519.

### DIFF
--- a/flit/buildapi.py
+++ b/flit/buildapi.py
@@ -1,6 +1,0 @@
-from warnings import warn
-warn('A package has specified `build-backend = "flit.buildapi"` and is being '
-     'built with Flit >= 3.10. This is likely to break in a future version. '
-     'Please change the backend to flit_core.buildapi, and/or specify a '
-     'maximum version of Flit.')
-from flit_core.buildapi import *


takluyver:
OK, I said I'd abandon this idea if anyone found a project using `flit.buildapi` with new style metadata, and then I did a quick [Github search](https://github.com/search?q=flit.buildapi+language%3ATOML++NOT+is%3Aarchived+NOT+is%3Afork&type=code&p=1), and immediately found a bunch of projects doing exactly that. Nothing super prominent, perhaps, but there's no point in breaking these for the sake of a few lines that don't need any maintenance.

<details>
<summary>List of example files found on Github (not exhaustive)</summary>

https://github.com/karthagokul/sentimeter/blob/main/pyproject.toml
https://github.com/gd3kr/cx-copilot/blob/main/pyproject.toml
https://github.com/microsoft/pytest-azure/blob/main/pyproject.toml
https://github.com/ocrosby/notes/blob/main/python/notebooks/marimo-nb/pyproject.toml
https://github.com/henu/zerodup/blob/master/pyproject.toml
https://github.com/osintbuddy/plugins/blob/main/pyproject.toml
https://github.com/shaneish/pyrchive_is/blob/master/pyproject.toml
https://github.com/pyasm-lib/pyasm/blob/main/pyproject.toml
https://github.com/jazielloureiro/LocalLeaf/blob/master/pyproject.toml  
https://github.com/jmasonlee/pyspark-dataframe-wrappers/blob/main/pyproject.toml
https://github.com/piderking/Board/blob/master/pyproject.toml
https://github.com/ignisf/sbm69/blob/master/pyproject.toml
https://github.com/meinardmueller/synctoolbox/blob/master/pyproject.toml
</details>

So, `flit.buildapi` stays, probably forever. You're still encouraged to use `flit_core.buildapi`, and [specify a version range](https://flit.pypa.io/en/stable/pyproject_toml.html#build-system-section) to avoid breaking on future major versions. But I'm not going to break stuff that would otherwise work just to do a tiny bit of cleanup.

PR #796, opened 2026-03-27
Pin actions and dependencies in release workflow

sethmlarson:
Hello! This pins both the GitHub Actions and the Python packages used during the release process. If you want we can also change the frequency of GitHub Actions updates, too. Right now it's daily, with pins maybe that will be overwhelming? Might make sense to change to less frequently.


.github/workflows/test.yml
```diff
@@ -15,6 +15,9 @@ concurrency:
     ${{ github.event.pull_request.number || github.sha }}
   cancel-in-progress: true
 
+permissions:
+  contents: read
```
cdce8p, line 19: This can also be set as default for the repo / org in which case this is kind of redundant here. I don't have access to the settings, so I can't check. It should be under `Actions` -> `General`

Image

.github/workflows/test.yml
```diff
@@ -60,26 +65,28 @@ jobs:
...
-        uses: actions/setup-python@v6
+        uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6
         with:
           python-version: "3.12"
 
       - name: Build flit_core & flit packages
         run: |
           python flit_core/build_dists.py
-          pip install requests docutils
+          python -m pip install -r packages-requirements.txt
```
cdce8p, line 80:
```suggestion
          python -m pip install -r .github/workflows/packages-requirements.txt
```
?

.github/workflows/test.yml
```diff
@@ -32,10 +35,12 @@ jobs:
           - "3.13"
           - "3.14"
     steps:
-      - uses: actions/checkout@v6
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          persist-credentials: false
```
cdce8p, line 40: What is the advantage of using `persist-credentials` really? The `GITHUB_TOKEN` is already available to all steps and is configured with read-only anyway.

.github/dependabot.yml
```diff
@@ -4,3 +4,9 @@ updates:
     directory: "/"
     schedule:
       interval: "daily"
```
cdce8p, line 6: I think it makes sense to adjust that as well, like you suggested.

cdce8p (commented):
Thanks for the PR @sethmlarson

.github/workflows/test.yml
takluyver: It's currently set to allow read & write. The only thing I can see that might require write access is `actions/upload-artifact`, though I'd guess that artifacts within the actions system are a separate thing and don't need write permission. I'm happy to try switching it to read only if that makes sense.

.github/dependabot.yml
```diff
@@ -4,3 +4,9 @@ updates:
...
+  - package-ecosystem: "pip"
+    directory: ".github/workflows/"
+    cooldown:
+      default-days: 7
+    schedule:
+      interval: "quarterly"
```
takluyver, line 12: I might step this up to monthly 🤷

.github/workflows/packages-requirements.txt
```diff
@@ -0,0 +1,153 @@
+#
+# This file is autogenerated by pip-compile with Python 3.13
+# by the following command:
+#
+#    pip-compile --generate-hashes --output-file=.github/workflows/packages-requirements.txt --pip-args='--only-binary=:all:' --strip-extras .github/workflows/packages-requirements.in
```
takluyver, line 5: Out of interest, if a new version of one of these packages has a new dependency of its own, is dependabot smart enough to add that to this file? Or even just flag that there's a new dependency that should be pinned?

.github/workflows/test.yml
cdce8p: I've set it to `Read` only for all my own projects and `upload-artifact` still seems to work fine. Based on https://github.com/actions/upload-artifact/issues/197 it seems the action uses some kind of custom token with it's own permission details.

I'd just suggest to try changing it here. We should see if it works during the next workflow run.

takluyver: Thanks, that makes sense. I've set it to read-only now, and unchecked "Allow GitHub Actions to create and approve pull requests"

PR #806, opened 2026-07-16
Fix wheel file permissions being too restrictive

### LINKED ISSUE #804: Wheels are created with too strict permissions
When I run `flit build` I get a wheel and a source tarball. The tarbar has permissions 0644 ('-rw-rw-r--'), which matches my umask. However, the wheel has permissions 0600 (`-rw-------`), which is (IMHO) to strict, as other users are not able to read the wheel.

Hatch had a [similar isse](https://github.com/pypa/hatch/issues/1519) and I guess flit has the same reasons behind this issue.

If the current permissions are by intention, please explain why and maybe document this somewhere (or point me to existing docs, which I couldn't find).

pctablet505:
Fixes #804

`make_wheel_in()` builds the wheel in a temporary file created by `tempfile.mkstemp()`, which sets the file's permissions to `0600`. That mode was never relaxed before the temp file was renamed into its final `.whl` path, so built wheels ended up owner-only readable instead of the usual world-readable `0644`. This breaks setups where the wheel is later read by a different user (e.g. some CI/build pipelines).

The fix normalizes the temp file's permissions with the existing `normalize_file_permissions()` helper right before it's renamed into place, matching the same normalization already applied to the entries inside the wheel. Added a regression test that checks the built wheel's mode on disk.

Same root cause as pypa/hatch#1519, fixed there with the analogous `normalize_artifact_permissions()`.

### DIFF
--- a/doc/history.rst
+++ b/doc/history.rst
@@ -1,6 +1,12 @@
 Release history
 ===============
 
+Version 4.0.3
+-------------
+
+- Fix built wheels being created with overly restrictive ``0600`` permissions
+  instead of the normal, world-readable ``0644`` (:ghpull:`806`).
+
 Version 4.0.2
 -------------
 
--- a/flit_core/flit_core/wheel.py
+++ b/flit_core/flit_core/wheel.py
@@ -222,6 +222,12 @@ def make_wheel_in(ini_path, wheel_directory, editable=False):
             wb = WheelBuilder.from_ini_path(ini_path, fp)
             wb.build(editable)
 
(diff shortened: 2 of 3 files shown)


takluyver:
Thanks!

Ideally the file should be created respecting the current umask, rather than with effectively fixed 644 permissions. But frustratingly, there isn't a simple API to *get* the current umask without modifying it. I think we could work around this by creating a temporary directory, then making a file inside it (which would get the default permissions). But I'm not sure that it's worth it.

PR #809, opened 2026-07-23
Support SPDX WITH expressions

### LINKED ISSUE #735: Support WITH in SPDX license expressions
A follow on from #723. PR #731 added support for compound license expressions using the AND/OR operators and parentheses, but WITH was left for later.

This will require adding the list of exception IDs from SPDX to the list of license IDs.

yangfan-yf-yf:
Closes #735.

This adds SPDX exception IDs for `WITH` expressions. The parser now normalises exception IDs, retains the existing uppercase-operator rule, and rejects a second `WITH` after an exception ID.

The generated license data remains aligned with SPDX v3.26, which matches the version already used for the existing license IDs; this change adds only exception IDs.

Validation:

- `python -m pytest flit_core/tests_core -q` with `PYTHONPATH=flit_core` (216 passed, 1 skipped)
- Boundary checks for case normalisation and repeated `WITH`


flit_core/flit_core/config.py
```diff
@@ -740,10 +746,13 @@ def normalise_compound_license_expr(s: str) -> str:
                     raise ConfigError(invalid_msg.format(s=s, reason=reason))
                 parts.append(part)
             else:
-                if parts and parts[-1] not in {' AND ', ' OR ', '('}:
+                if parts and parts[-1] not in {' AND ', ' OR ', ' WITH ', '('}:
                     reason = "a license ID must follow either AND, OR, or '('"
```
takluyver: The message here should be updated now that WITH and exception IDs are also accepted

takluyver (commented):
Thanks, this looks good to me.

Could you add an entry in the documentation about version constraints, that license expressions using WITH require `flit_core >=4.1`?

yangfan-yf-yf:
Updated in 33a7a39:

- Added the `flit_core >=4.1` requirement for license expressions using `WITH` to the version constraints documentation.
- Updated the parser error message to mention exception IDs and `WITH`.

Validation: core tests passed (217 passed, 1 skipped), the valid-expression and updated-error-message checks passed, and the Sphinx HTML build passed with warnings treated as errors.

PR #824, opened 2026-09-14
install: keep explicit CRLF in Windows .cmd wrappers

r3wretrhy:
Windows console wrappers are written with explicit `\r\n`. Opening the `.cmd` file in text mode still translates `\n` to the platform newline, so the file on disk becomes `\r\r\n`.

Open with `newline=''` (and UTF-8, matching the neighbouring Unix script) so the CRLF that is already in the string is what gets written.

The added assertion in `test_install_package` fails on an unpatched Windows tree (`\r\r\n` in the bytes) and passes with the change.


takluyver:
Thanks

takluyver:
Thanks

PR #825, opened 2026-09-15
doc: list all four fields accepted in project.dynamic

hxperl:
`doc/pyproject_toml.rst` says `dynamic` accepts only `"version"` and `"description"`. It has accepted four since 4fdd194a (PEP 794 support), which widened `allowed_dynamic_fields` without touching the docs.

With `dynamic = ["version", "description", "import-names"]`:

```
>>> read_flit_config(Path("pyproject.toml")).dynamic_metadata
['description', 'import-names', 'version']
```

And flit's own error already lists all four:

```
ConfigError: flit only supports dynamic metadata for:description, import-names, import-namespaces, version
```

Doc line updated to match. Sphinx builds clean with `-W`; `flit_core/tests_core` 218 passed, 1 skipped on macOS arm64 / CPython 3.14.


takluyver:
Thanks

